PR #97, opened 2015-05-31
Feature/arquillian infrastructure

yersan:
This pulll request includes the resolution of issue-21 together with an integration test for that issue.
Maybe it could be an initial point to bring arquillian tests to BootsFaces. Currently it is just testing that the rendered markup is according with the .xhtml file.
I have used firefox and phantom.js as test browsers with a 3.2 glassfish embeded which included by default mojarra 2.1.6
To run the test: mvn clean verify -Pglassfish-embedded
Waiting for feedback.


stephanrauh:
I've accepted and merge the pull request. ¡Muchas gracias!, ¡Yeray!

PR #119, opened 2015-06-28
Feature/unit testing

yersan:
It seems that there are some class path libraries pollution with wildfly-managed profile running unit and integration tests. Currently it is not possible execute the unit testing in this profile. I haven't investigated in deep the root cause but at least with this change it is possible the execution of Integration tests using wildfly-managed profile. This change disables the unit tests this profile.

With this PR, if we need to run unit tests:
mvn clean test

To run integration test, as we used to do:
mvn clean verify -Pwildfly-managed

This pull request includes an additional package name change, all integration tests will be under: 
net.bootsfaces.it.*

And all unit tests under:
net.bootsfaces.ut.*

I hope that you can find it useful.


stephanrauh:
I've accecpted and merged the pull request.

Issue #126, opened 2015-07-09
Bug: b:panel collapsible=false


jottyfan:
regarding bootsfaces 0.7.0 from maven central repository

on using <b:panel collapsible="false" ..., the renderer renders a closing </a> tag from the collapsible=true configuration. This comes up by myfaces in a warning: HTML nesting warning on closing a: element h4 rendered by component...

Solution: fix rendering </a> only when collapsible != false

jottyfan:
This is found in net.bootsfaces.component.panel.PanelRenderer line 162:

replace rw.endElement(H.A); by

if (isCollapsible) {
  rw.endElement(H.A);
}

Could anyone with write access please add this fix? Thanks!

stephanrauh:
Done. Thanks for pointing this out!

jottyfan:
Created a pull request for 0.8.0 that contains that bug even longer

stephanrauh:
Oops - did I forget to commit my bugfix? In any case: thanks for your pull request!

Regards,
Stephan

PR #230, opened 2016-01-05
bug fix: close open a tags only

jottyfan:
When collabsible is false, the not opened a tag should also not be closed.


stephanrauh:
See issue #126

PR #264, opened 2016-01-20
misleading exception

jottyfan:
exception leads to wrong file that never exists, therefore, please approve this little bug fix


stephanrauh:
Thanks!

PR #281, opened 2016-02-02
Issue #279 - Persistent DataTable properties attribute.

### LINKED ISSUE #279: DataTable option that allows the state of the table to last between refreshes / re-rendering
I thought of this while working with a DataTable that had a column with buttons in it that allowed me to re-order the items in the backing list, then re-render the table via AJAX. The problem comes when you're re-ordering items on pages other than the 1st page. When you re-render the DataTable, you start back on page one.

It would be nice to have an attribute for the DataTable that allows you to bind the state of of the component (page, search input, currently sorted column, results per page dropdown, etc) to a backing bean. This would allow the table properties to inherit the scope of the backing bean (`@SessionScoped`, `@ViewScoped`, etc).

I haven't looked through the actual BootsFaces code for this component but here's how I imagine it would work:

``` java
@ViewScoped
@Named
public class MyBackingBean implements Serializable {
    private List<SomeObject> someObjectList;
    private Map<String, Object> dataTableProperties;

    @PostConstruct
    void postConstruct() {
        someObjectList = SomeService.gimmeAList();
    }    

    public List<String> getSomeObjectList() {
        return someObjectList;
    }    

    public Map<String, Object> getDataTableProperties(){
        return dataTableProperties;
    }

    public void setDataTableProperties(Map<String, Object> dataTableProperties){
        this.dataTableProperties = dataTableProperties;
    }
}
```

``` xml
<b:dataTable
    value="#{myBackingBean.someObjectList}"
    var="obj"
    dataTableProperties="#{myBackingBean.dataTableProperties}>
    ....
    ....
</b:dataTable>
```

If the attribute is not null in the DataTable component, then it would read / write to that map for all of it's properties :)

Hope I was clear.

XaeroDegreaz:
fix #279

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,5 @@ src-gen
 /gradleResources/.nb-gradle/
 .DS_Store
 .gradle
+.idea/
+*.iml
--- a/src/main/java/net/bootsfaces/component/dataTable/DataTable.java
+++ b/src/main/java/net/bootsfaces/component/dataTable/DataTable.java
@@ -55,7 +55,7 @@ public class DataTable extends UIData implements IAJAXComponent, ClientBehaviorH
 
 	private Map<String, Object> attributes;
 
-    protected enum DataTablePropertyType
+    public enum DataTablePropertyType
     {
         currentPage
     }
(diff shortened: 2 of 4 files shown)


XaeroDegreaz:
The taglib still needs to be updated and other documentation. What is the process you dudes normally take to generate that or is it something you do by hand?

stephanrauh:
Look at the `bootsfaces.jsfdsl` file in the XText folder. Add the attribute there. After that, we generate the taglib fragments and add them manually to the taglib. I hope we can automate this step soon. 

If you want to use the generator yourself: it's an Eclipse plugin hosted at https://github.com/stephanrauh/JSFLibraryGenerator.

Issue #288, opened 2016-02-10
datepicker doesn't use converter


lyahim:
Hello All,

I use in backing bean LocalDate and try to inject into datepicker component.
When the page is loading the component said: Value could be either String or java.util.Date.
I use `<f:converter ... />` the component shouldn't use the converter?

Best Regards,
Mihaly

zhedar:
In #341 I stumbled upon the Date and Time bits in the `DatePicker`. I'm a little concerned about the changes you did in https://github.com/TheCoder4eu/BootsFaces-OSP/pull/292 and hope you can shed a little light on how using Java8 in Java6 source can work.
What baffles me is, that Maven doesn't complain about that at all. I thought it should throw a compile errors at least if the source and target level is set to 1.6.
However I'm running it with JDK8, so it may just run fine. But couldn't this change break Java6 environments? Did anyone ever run the Snapshot in such an environment and did test that?

It may be convenient to be able to use the new Date and Time API directly with the component, but the standard way still seems to be to [define a custom converter](http://jj-blogger.blogspot.de/2015/06/utilizing-java-8-date-time-api-with-jsf.html) to keep the 1.6 compatibility of the framework.

**TL;DR:** We should check for 1.6 compatibility in the current snapshot and probably remove the `LocalDate` parts again.

Issue #144, opened 2015-07-28
Missing attributes in `<b:datepicker`


AlexanderBl4h:
According to the PrimeFaces `<p:calendar>` (http://www.primefaces.org/showcase/ui/input/calendar.xhtml) there are in my opinion some missing attributes in the BootsFaces `<b:datepicker>`:
- required: always usefull (I mean, why not?)  and message.
- min/max date
- pattern 
- disabled

Also the time is missing. Maybe it is better to use a calendar instead of a datepicker and create an attribute like `time` which can set to `false` or `true` (true default) to display/use the time.

stephanrauh:
As for the time picker: this one might be nice: http://www.daterangepicker.com/. It also covers date and time ranges.

AlexanderBl4h:
:+1:

stephanrauh:
The datetimepicker seems to be popular - see #148. We should also think about supporting the Joda date time library and the new date / time classes of Java 8.

anemdhana:
Thanks and using datetimepicker than daterangepicker ? See if you think these converter classes are useful to the project. https://gist.github.com/anemdhana/403c0923c1e25ff289fe

stephanrauh:
@anemdhana Yes, your classes look useful. Thanks!

stephanrauh:
As it turns out, `disabled`, `min`, `max` and `pattern` are already there. We just forgot to document them and to add them to the taglib (i.e. to your editors autocomplete list).

AlexanderBl4h:
How does the pattern work? I can not figured it out. :( Can you please provide some examples?

zhedar:
@AlexanderBl4h You can supply a pattern by setting the `dateFormat` attribute. However, while testing this I found, that it only works, if `lang` is set as well. I probably have a closer look at that, because it seems redundant. The pattern itself uses the [SimpleDateFormat patterns](https://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html).
A rather exotic example may be:

    <b:datepicker lang="it"  dateFormat="dd,MM"/>

I noticed, that setting the lang code to `en` won't apply the supplied  `dateFormat`. Other codes like `de`, `it`, `es` seem to work though. That's definitely something we still need to work on.

stephanrauh:
@zhedar I've assigned the ticket to you, because you seem to have started investigating it.

zhedar:
[...]

PR #292, opened 2016-02-16
#144 #288

lyahim:
- For #288 cannot solve to usage the of converter, just add support to java8 features
- add hu locale to datepicker


TheCoder4eu:
Thanks! I just merged your PR.

PR #304, opened 2016-02-22
Bug fix and changes to build.gradle

asterd:
- Added support for multi-theme gradle compilation: with this change, you can provide a list of comma separated theme names and the build system generates resources for all listed themes. A custom behaviour is managed for "other" with the simple copy of additional css.
- ScrollUp component: added new attributes to match the original library ones. For documentation: i have created the xhtml file with Stephan eclipse plugin.. where do i have to put it?
- Bug-Fix to navBar for 'non-sticky' attribute: here is the big edit. A deeper test of the old push request in another project, has made me find a bug: the html-related css not coming with the attribute. And, with a custom theme, it loses the non-sticky functionality. So i have modified NavBar to manage a PostAddToViewEvent and load correct custom css file only if the attribute is specified inside the navbar tag.


stephanrauh:
Great, thanks! It doesn't even look difficult... you've approached the multi-theme-build from a slightly different angle than I did, and all of a sudden things become simple :).

As the the "non-sticky"-patch: +1!

stephanrauh:
Something seems to be wrong - possibly on my machine. When I run "gradle clean" and "gradle assemble", the build finishes within 3 seconds. Last time I tried, it took 17 seconds, and a couple of file in the folder "mavenResources" were modified. What do I do wrong?

asterd:
Hum.. Sounds strange.. Tomorrow i'll try on my machine to see if i'm doing something word.. Only one hint... In the merged file list i see also a gradle.prefs that is configured for my personal environment.. I don't know why git plugin doesn't ignore this file... But i think is better to remove it...

asterd:
I have tested the build.gradle.. if i run the clean and assemble tasks also to me takes about 3 seconds to run and it produces a 380k jar...  Sincerely i don't understand why it doesn't run the bootswatch download  task (i think this is the problem).. This is what i get:

schermata 2016-02-23 alle 09 25 59

schermata 2016-02-23 alle 09 28 32

It seems to run the addBootswatchTheme task but it does nothing. I don't think that this behavior is related to my edit, because my new tasks have to run after the download task. I have notice this behavior already in the past, so i run always the default task of Bootsfaces-OSP build.gradle and all works ok..

stephanrauh:
Running the target "buildResources" now works for me. The "assemble" tasks tries to optimize the build and to avoid unnecessary builds - which is bad in our case :).

asterd:
Very good!:) 
Another question: have you a roadmap to next release (maybe 0.9)? I mean approximate timing... Meanwhile i'm trying to implements some other components but is a  new world for me so i have to understand well all the behavious!

stephanrauh:
@asterd I just saw that the loop doesn't seem to work correctly. After running the gradle build, the themes in the folder mavenResources/META-INF/resources/bsf/css are identical. It seems like the theme is not generated again, but copied to the other theme folder. Would you mind to have a look?

stephanrauh:
Originally, we wanted to release a 0.9 or even a 1.0 during the JavaLand conference (March 08, 2016). However, this doesn't seem realistic, given how little time we've left. In general, my goal is to release three to four times a year. Which is to say, we manage to publish a release two or three times a year :). We haven't talked about the next release date yet, but I guess it might be in April.

asterd:
Ok..i will look at this problem tomorrow.. Ok also for the planning! Thank you!

asterd:
Hi, i have looked at the problem of the build.gradle. Basically, the problem is related to the lesscss task that calls a sub-build that i can't transform to support multiple generation. I'm sorry, i was too optimistic. So, there is no way (for now) to run a gradle task that builds all themes. But, i don't like to run 17 times the build script, so i wrote a simple java project that use the gradle-tooling-api makes possible to automate the execution of a gradle task, passing different arguments. In other words, this simple java class call the default task of the build.gradle, in a loop through a list of enabled themes. So, the build is done x times.. but is the program that does the work! To obtain this, i have modified the build script to get (if specified), a command line argument. Furthermore, i have also edit the build to prevent the deletion of the bootstrap e bootswatch zip to prevent multiple downloads. So, now, the entire build process tooks 4-5 sec per theme. The project is on my git at this address: https://github.com/asterd/BuildBootsFaces.git. The edit to the build.script is not committed on git because i'm working on a very basic and simple implementation of a treeview component and i want to end this development before committing it.

stephanrauh:
+1! I didn't have the time to look into your project today, but you definitely made my day!

PR #313, opened 2016-02-26
New components and various fix

asterd:
Hi, this is a very big pull request.
In this request there is:

1. Fix bug in multi-theme compilation. Now it support an external parameter for theme selection per single build. This is done to enable execution of gradle build from an external project (https://github.com/asterd/BuildBootsFaces.git) 
2. Added 3 new components: DefaultCommand, Tree and ColorPicker. The tree component uses a listener in the same way that primefaces does. To test these new components, i have created a very simple eclipse project (with all libs already inside), to understand how components work. If you want to try before commit, you can download this project and try it. The bootsfaces.jar inside is already the one modified. The project is located here: https://github.com/asterd/BootFacesTest.git
NB. The ColorPicker component is basically a copy of InputText field that adds some js and css and a simple js init configuration. I have not edit the original InputText for two reasons: 
  - The component is not mine and i don't know if you have some other ideas about it's future evolution
  - If you plan to add something like to masking or other things like that, these features does not apply to colorPicker, so i tought that is a good idea to mantain a drastic separation between the two components. But if you prefer to integrate this functionality to the original InputText there is no problems.
3. Add some utility methods: i have also add some simple methods to FacesMessages utility class.

Only one thing: i cannot test this release on IE (because i have not a IE7-8) so maybe this need a little more testing.. but in most cases I'm using jquery plugins that seems to work very well..


stephanrauh:
I've merged the pull request into the master branch.
ToDos left:
* add the new components to the BootsFaces.jsfdsl file
* verify if add snake-case attributes need to be added
* add documentation pages
* review the implementation
* move the `decode` method to the renderer
* check the license of the JS files of the new components

asterd:
Hi Stefan, i have two questions related to your ToDos:
1. what do you mean with snake-case attribute? i see that no tags uses the snake-case... is a future improvement?
2.  how can i add the unified ajax engine on b:tree? the ajax call that i use is directly in generated javascript and i didn't see any method inside AjaxRenderer class that can help me.. I'm loosing something?

stephanrauh:
1. Most Java-Programmers prefer the camelCase for variable names. On the other hand, most web designers seems to prefer snake case - i.e. putting a "-" between word boundary. For example, AngularJS uses snake-case in the HTML markup (like "ng-hide") and camel case in the back end ("ngHide"). Some time ago, I've decided to support both with BootsFaces. We haven't converted every component yet, but in many case we accept both "style-class" and "styleClass". The *.jsfdsl file make it easy to support both variants.

2. I've designed the unified AJAX engine to support every JavaScript callback, such as onClick, onMouseOver and so on. Unfortunately, I didn't document it yet, so I admit it's hard to figure out how to use it. Actually, I'm not entirely sure if the tree component can be used with the AJAX engine. But it would be nice if it was possible, so I've decided to continue the research one day or another. Mind you: if the AJAX engine supports b:tabView, it should be able to support b:tree too - both implementations are similar.

asterd:
Ok! Got it! With these information i can do something better for the future! :) by the way i'll look at the tabview to see the use of ajax... In the next pull request i try to add also the snake-case!

asterd:
I have only one question: i've tried to integrate the snake-case attribute but seems that AttributeMapWrapper does not manage it correctly, so the snake-case attribute does not work. The base problem is that it raise an IllegalArgumentException if the value is different from "String". This can be bypassed hard-coding a type converter but i don't like this way. And, even with this method, if the attribute map an EL expression, it doesn't work at all.. How can you think to manage that cases? I've seen that one way can be the extension of a StateHelper but Mojarra and MyFaces has differents stateHelper implementations....

stephanrauh:
It seems I've encountered the same problem already. That's why there an extra if for the `tooltip-delay` (which is an integer) and the `closeOnEscape` attribute (which is a boolean). You're right, on the long run we should opt for more generic solution. The generic method `convertValueToType` was a first (failed) attempt to do just that.

I didn't encounter the EL problem yet, but maybe that just due to sloppy testing. I suppose that's a simple extension of the `AttributeMapWrapper`: if the `value` is a `ValueExpression` or a `MethodExpression`, just put it into the hash map without converting it to a string. Did you try this approach already?

asterd:
This is not the already default behaviour? Anyway, if i put the value inside the map 'as is' the el expression is not managed and the system doesn't work (i have tried with the event listener of the tree)...

stephanrauh:
Oh, really? That's a nasty surprise. I hoped to be able to simply copy or move it from one key to another.

asterd:
Hi Stephan, 

i've looked at the problem of snake-case attributes and after a lot of tries i've found a possible solution. 
First of all there is no need of the AttributeMapWrapper. The problem is separated in two section:
1. Mapping of attribute with basic java types (String, Boolean, Integer)
2. Mapping of attribute that accept EL expressions.

Reading the documentation, seems that this two type of attribute are managed in a different way. For the first one type, the system calls the getAttribute() method of UIComponent, for the second one it call setValueExpression instead, bypassing the getAttribute method. 

But let me analyze what i've done for this two types:
For the first problem (getAttribute), a solution can be your AttributeMapWrapper, but inside there we cannot know the type of the attribute. To bypass this problem, we can use the BeanInfo architecture. The jsf implementation uses the Inspector class of the jre to find the properties of the single tag-bean. But the snake-case attribute does not affect to any real method. What we have to do is to instruct the virtual machine to find this properties and bind to a different method. This is my implementation for the Tree component:

```
import java.beans.BeanInfo;
import java.beans.IntrospectionException;
import java.beans.Introspector;
import java.beans.PropertyDescriptor;
import java.util.ArrayList;
import java.util.Arrays;
import java.util.List;

public class TreeBeanInfo extends java.beans.SimpleBeanInfo {

	@Override
	public PropertyDescriptor[] getPropertyDescriptors() {
		// get properties from default property descriptors
		List<PropertyDescriptor> pdl = new ArrayList<PropertyDescriptor>();
		PropertyDescriptor rv[] = super.getPropertyDescriptors();
		if(rv != null) {
			pdl.addAll(Arrays.asList(rv));
		}

		try {
			// if i have found nothing, i'm trying to get manually
			if(pdl.size() <= 0) {
				BeanInfo bi = Introspector.getBeanInfo(Tree.class, Introspector.IGNORE_ALL_BEANINFO);
				PropertyDescriptor birv[] = bi.getPropertyDescriptors();
				if(birv != null) {
					pdl.addAll(Arrays.asList(birv));
				}
			}
			
			// then add custom properties
			// render-root
			PropertyDescriptor _renderRoot = new PropertyDescriptor("render-root", Tree.class, "isRenderRoot", "setRenderRoot");
			_renderRoot.setBound(true);
			pdl.add(_renderRoot);
			
			// node-selection-listener
			PropertyDescriptor _nodeSelectionListener = new PropertyDescriptor("node-selection-listener", Tree.class, "getNodeSelectionListener", "setNodeSelectionListener");
			_nodeSelectionListener.setBound(true);
			pdl.add(_nodeSelectionListener);
		} catch (IntrospectionException e) {
			e.printStackTrace();
			throw new Error(e.toString());
		}
		
		PropertyDescriptor[] array = new PropertyDescriptor[pdl.size()];
		return pdl.toArray(array);
	}  
}
```

With this class, the system found the snake-case attribute and call the camel case getter method (it works very well with render-root attribute in the example).

But this solution not work for the "node-selection-listener" attribute that uses a EL expression binding. So, for the second point, we have to override the setValueExpression method of the component, for provide a translation of the key from snake-case to camel-case. This is my implementation in Tree component (but can be used in any component without any modification):


```
	@Override
	public void setValueExpression(String name, ValueExpression binding) {
		if(name.contains("-")) {
			StringBuilder newKey = new StringBuilder(name.length());
			boolean toUpperCase = false;
			for (char c : name.toCharArray()) {
				if (c == '-')
					toUpperCase = true;
				else {
					if (toUpperCase) {
						toUpperCase = false;
						c = Character.toUpperCase(c);
					}
					newKey.append(c);
				}
			}
			name = newKey.toString();
		}
		super.setValueExpression(name, binding);
	}
```

In the end, with the override of setValueExpression and the definition of a BeanInfo class, we can manage all attribute in camelCase and snake-case. 
Obviously, if you like this solution and you think that can be applicable, it can be optimized with the creation a generic BsfUIComponentBase class that already override the setValueExpression and has a BeanInfo class that reads the custom property mapping with an abstract method (maybe.. i have not tested)... 

Let me know what you think about that!

Issue #254, opened 2016-01-15
DataTable attribute for default sorting column


XaeroDegreaz:

XaeroDegreaz:
Some people at work were asking again about being able to sort things on a specific column when the page loads, so i thought I'd go ahead and fix this one myself. PR is up and ready to merge at your whim :)

stephanrauh:
By the look of it, we've integrated your feature but forgot to close this issue. Be that as it may: default sort order is an official feature of BootsFaces 0.9.0 now.

Kokscola:
Can you give an example how to set a default sorting column?

PR #338, opened 2016-04-01
#254 - Enable user to specify default column sorting for DataTables

XaeroDegreaz:
#254


stephanrauh:
Looks like you're sorting the table on the client side? Nice! And thank you very much!

XaeroDegreaz:
You're welcome :) Yeah just using the JS DataTable's API to set column sorting when the table gets initialized.

PR #339, opened 2016-04-04
removing ugly positioning

jottyfan:
whenever css class form-control is used, it add a margin-bottom to the tag. This looks ugly when surrounded by text lines as here; therefore, please accept my change request to make the paginator dropdown look more user friendly.


stephanrauh:
Seems to be OK to me. I'll merge your request. Thanks!

Issue #220, opened 2015-12-28
Refreshing a region containing tooltips causes a memory leak


stephanrauh:
This problem can be observed in the AJAX+RTL demo of switch.xhtml in the showcase. Moving the mouse over the switch makes the tooltip occur. When you toggle the switch, it is replaced with a newer version of the same switch. However, the tooltip remains, and there's no way to hide it again. Looking at the end of the HTML view reveals that there are two tooltip divs. Each time you toggle the switch, another tooltip div appears.

stephanrauh:
Destroying the tooltip before sending the AJAX request might help: http://stackoverflow.com/questions/28220597/destroy-all-bootstrap-tooltips-in-a-div

stephanrauh:
Similar memory leaks should affect every component based on jQuery. Marco Rinck's JUA framework might help (after adapting it to jQuery instead of AngularJS).

stephanrauh:
Todo: update layout/navigationAndAJAX.jsf of the showcase after fixing the bug.

Issue #325, opened 2016-03-11
Unable to save dynamic action with clientId 'j_id...'


rpinquie:
Hi all,

I am getting many warnings as this one "Unable to save dynamic action with clientId 'j_id31' because the UIComponent cannot be found".

Could anyone explain me how to upgrade to OmniFaces 2.2? I do not have explictly specified any OmniFaces dependency, thus I am wondering which one I should modify.

Tanks in advance,

Romain

farooq7124:
Hi,

I am facing same issue with Prime faces. Please update any solution if anyone resolved.

Thanks,
Mohammed Farooq

asterd:
Hi,
can you provide an example that raise this issue? i'm trying to replicate it but without success.. 
About OmniFaces, if you use maven to build your project you have to change the pom.xml to referencing the newest version of OmniFaces, otherwise you can just download the newest jar and put inside your WEB-INF/lib...

stephanrauh:
@rpinquie It's unlikely that you get the error message without activating OmniFaces. Which libraries do you use? Is OmniFaces part of your WEB-INF/lib folder?

stephanrauh:
@farooq7124 This is a warning, not an error. As soon as you set the ProjectStage to production, the warning is gone. So I consider this a minor nuisance, not a serious error. In any case, do you get the warning without using OmniFaces?

rpinquie:
@stephanrauh Actually, OmniFaces is not part of my WEB-INF/lib folder. I was just investigating OmniFaces's solution as you advised me in a previous post.

Before moving to BootsFaces I was using PrimeFaces and never got such warning. Do you think it is maybe be due to the blend of BootsFaces and PrimeFaces?

stephanrauh:
@rpinquie Unlikely. The warning occurs when you manipulate the JSF DOM tree using the wrong API. For instance, by removing a resource file by simply deleting it from the list. That's what OmniFaces used to do in an earlier version. JSF detects that something's missing, gets nervous and reacts with omitting these confusing warnings. It's possible that BootsFaces uses the "forbidden" API, but I believed to have fixed that. In any case, can you upload a simple project that reproduces the warning? This make our job a lot easier.

stephanrauh:
I managed to reproduce the issue with this JSF view:
```xml
<?xml version='1.0' encoding='UTF-8' ?>
<!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/xhtml"
[...]

Issue #323, opened 2016-03-10
tooltip stills after poll


VonTalavang:
when a tooltip is activated...
and the form is refreshed...
the tooltip still... forever... and ever...

VonTalavang:
Yes, i guess is because the object is still waiting for the mouse event to remove the tooltip... and if the trigger dissapear... so the message stills there. 
Even if i turn of the monitor !!... (drum joke sound)

zhedar:
@VonTalavang  I currently can't reproduce this. Could you provide a minimal snippet?
Here's what I have tried:

    <h:form>
        <b:button value="Hover over me" tooltip="What a nice tooltip!" tooltip-position="auto"/>
       	<b:poll update="@form" />
    </h:form>

Hovering over the button creates the Tooltip, but it doesn't stay after some updates. Also which browser were you using for this to happen? I've encountered several weird tooltip errors, where the browser was to blame.

VonTalavang:
chrome.

I remove the tooltips. i dont have code available right now.

zhedar:
Thank you for your response.
I've been able to reproduce this in Chrome and Firefox. Safari however doesn't have this issue, that's why I couldn't reproduce it in the first place.

VonTalavang:
you are welcome !
thanks for the support

zhedar:
This issue may be related to https://github.com/twbs/bootstrap/issues/3084, which recommends to manually hide the tooltip on ajax actions (which remove components, which in turn could be triggered by JSF in this case). My knowledge of JSF's internal ajax behaviour isn't that great yet, so maybe I'm on the wrong path there.

However, manually looking for tooltips and destroying them on ajax requests isn't an option in my opinion, it's just too much overhead in case of frequent ajax requests for such a rare use case. Also you would hide tooltip immediately on every ajax request, which is a negative side effect.

Maybe there might be other hacks for this, that would be applicable though (see referenced bootstrap issue). 
Any thought on that?

stephanrauh:
Yes, the problem is a fairly common problem with AJAX request. It includes a lot of components - tooltips, popovers, maybe even modal dialogs. Sometimes you don't even need AJAX to run into problems. A simple tooltip on each row of a datatable causes problems if the datatable has a paginator, sort buttons or a "search" field. I'd suggest to simply hide every popup and every tooltip on every AJAX request.

stephanrauh:
Also see #220. Please close #220 when you close this issue.

[...]

PR #347, opened 2016-04-17
Refactor AddResourcesListener and Tooltips

zhedar:
The AddResourcesListener currently is a bit long and contains a lot of boilerplate code, therefore I made some refactorings. ~~However, **this is still WIP, so probably shouldn't be merged yet**. For example `addJavascript` is still too long (220 lines for a single method) and the `Comparator` could be moved into an own class.~~

~~I just hoped to get some feedback (since it's getting late), e.g. on the difference between `addResourceToHeadButAfterJQuery` and `addBasicJSResource`. I couldn't spot any other than the different resource keys, but somehow I couldn't just get it to run with my extracted `addResource` method. The first seems to fail whereby the latter doesn't, so I just kept `addResourceToHeadButAfterJQuery` in place until I found the difference.~~

Changes:
* made yes / no checks more readable and fixed case errors
* extracted methods to reduce boilerplate code
* reduced nesting levels
* fixed typos

### DIFF
--- a/src/main/java/net/bootsfaces/C.java
+++ b/src/main/java/net/bootsfaces/C.java
@@ -40,7 +40,7 @@ public final class C {
     
     //Font Awesome
     //public static final String P_USEFONTAWESOME ="BootsFaces_USE_FA";
-    public static final String FA_VERSION="4.5.0";
+    public static final String FA_VERSION="4.6.1";
     public static final String FONTAWESOME_CDN_URL="//maxcdn.bootstrapcdn.com/font-awesome/"+FA_VERSION+"/css/font-awesome.min.css";
     
     //Meta tags
--- a/src/main/java/net/bootsfaces/component/GenContainerDiv.java
+++ b/src/main/java/net/bootsfaces/component/GenContainerDiv.java
@@ -79,7 +79,7 @@ public class GenContainerDiv extends UIComponentBase {
         }
         context.getResponseWriter()
                .endElement("div");
-		Tooltip.activateTooltips(context, getAttributes(), this);
+		Tooltip.activateTooltips(context, this);
 
     }
 
--- a/src/main/java/net/bootsfaces/component/alert/Alert.java
+++ b/src/main/java/net/bootsfaces/component/alert/Alert.java
@@ -40,10 +40,9 @@ public class Alert extends UIComponentBase implements net.bootsfaces.render.IHas
 	public static final String DEFAULT_RENDERER = "net.bootsfaces.component.alert.Alert";
 
 	public Alert() {
-		Tooltip.addResourceFile();
+		Tooltip.addResourceFiles();
 		AddResourcesListener.addThemedCSSResource("core.css");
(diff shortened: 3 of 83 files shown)


zhedar:
I just added another refactoring to this PR.
This one changes the way, how Tooltip resources are added (always add tooltip.css and tooltip.js in one method call).

However, in retrospect I saw that in some cases `Tooltip.addResourceFile()` wasn't called, although the `tooltip.css` file was added. Examples for this are `Message`, `Badge` or `Messages`. In other cases, only the js file was added, e.g. in `CarouselControl` or `CarouselItem`.
Is there a specific cause why these components just need one of the files or was it just forgotten to add the other one?

zhedar:
I just observed, that `selectOneMenu`'s tooltip isn't shown, maybe the data attributes are not set properly. This problem is already present in the [current snapshot showcase](http://www3.bootsfaces.net/Showcase/bootstrap/tooltips.jsf), so not a breaking change from my most recent changes. However, it isn't present in the [0.8.1 Showcase](http://showcase.bootsfaces.net/bootstrap/tooltips.jsf), so probably a bug we introduced in the Snapshot itself.

zhedar:
I just tried to add an ajax listener to remove the tooltips to fix #220 and #323, but somehow this got snowballed with every ajax request, if it was added in `Tooltip.activeTooltips`.
It seems like `Tooltip.activeTooltips` is called on any ajax request as well, so I just added a quick `$('.tooltip').tooltip('destroy');` to its script.

While fixing this I discovered strange behaviour of `b:poll`, which polled just way too fast and saw #325 a lot again, I may investigate on that later on.

zhedar:
The last commit I added(49e4a74) fixed the #325 related bug, which happened because PrimeFaces components seem to be duplicated.
The other warnings I saw came from the fact, that I used Bootstrap from CDN accidentally. We seem to remove every bsf related resource on such an occasion, which doesn't look like correct behaviour and should be fixed in the future.

I also solved the problem from the entry post and could replace two similar addResource methods by a generic one.

There's only one TODO left, before I consider this PR complete: There's still a nested if statement I marked as TODO in 49e4a74 I'd like to combine, but I'm not sure if that is actually correct as it stands, because it looks like the second statement always yields true if the first did as well.

zhedar:
I just cleared the last open point on my list, but I would like someone to review my latest change (e6e4e46). @stephanrauh maybe? The old statement looked kind of odd. It seems like it always evaluates to true anyways, what is this needed for? Do we even need those checks?

Otherwise I would declare this PR done... before it get's an big ugly abomination of changes I stitched together :O

stephanrauh:
Whow! That's a huge one! I'm starting to test and merge this PR.

zhedar:
It would look smaller, if I wouldn't have named all the little changes I did and put them into separate commits though. I've tested most of this for two weeks now, so I'm rather confident that everything is going to be ok ;) Thanks for testing!

PR #351, opened 2016-04-21
Don't let @previous loop endlessly if there is no previous component

### LINKED ISSUE #256: @previous doesn't work if it's in a b:column
This code doesn't work. That's obvious, but it seems to result in a endless loop.
```xml
<b:column>
   <b:message for='@previous' />
</b:column>
```

zhedar:
Due to #256, the `@previous` expression did produce an endless loop, if there was no component to be found on the same level.

This should handle that, as well as document what happens in the crucial part. The problem was, that there was no break, if i was actually below 0. In this case it just ran forever.

@stephanrauh Could you test this? I only ran very limited tests, because I usually doesn't use that expression.

This fixes #256 .


stephanrauh:
Actually, I didn't start working on #256 yet :). But thanks for the bug fix!

PR #401, opened 2016-06-08
Added info to maven generated MANIFEST.MF so that maven generates a similar MANIFEST.MF to gradle.

stiemannkj1:
[The Gradle generated BootsFaces jar](http://www.bootsfaces.net/download/BootsFaces-OSP-0.8.6-dist-default.jar) contains a lot of useful info in the **`MANIFEST.MF`**. Unfortunately, [the Maven generated BootsFaces jar](http://central.maven.org/maven2/net/bootsfaces/bootsfaces/0.8.6/bootsfaces-0.8.6.jar) does not contain the same info in the **`MANIFEST.MF`**. This PR adds the info that is missing from the Maven generated **`MANIFEST.MF`**. The **`MANIFEST.MF`** files still have some minor differences, but the important information is the same. Please consider merging this PR in order to reduce the differences between the two generated jars and to provide the same info to Maven developers.

Thanks,
\- Kyle

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -44,6 +44,7 @@
 	<properties>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<project.reporting.outputEncoding>${project.build.sourceEncoding}</project.reporting.outputEncoding>
+		<maven.build.timestamp.format>yyyyMMdd-HH:mm:ss</maven.build.timestamp.format>
 
 		<!-- compiler plugin configuation -->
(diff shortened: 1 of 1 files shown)


stiemannkj1:
I just read the contributing guidelines. Re-sent as https://github.com/TheCoder4eu/BootsFaces-OSP/pull/402.

stephanrauh:
@stiemannkj1 Reading and following our coding guidelines is very attentive of you. Thank you very much!

PR #402, opened 2016-06-08
add info to maven generated MANIFEST.MF so that maven generates a similar MANIFEST.MF to gradle

stiemannkj1:
[The Gradle generated BootsFaces jar](http://www.bootsfaces.net/download/BootsFaces-OSP-0.8.6-dist-default.jar) contains a lot of useful info in the **`MANIFEST.MF`**. Unfortunately, [the Maven generated BootsFaces jar](http://central.maven.org/maven2/net/bootsfaces/bootsfaces/0.8.6/bootsfaces-0.8.6.jar) does not contain the same info in the **`MANIFEST.MF`**. This PR adds the info that is missing from the Maven generated **`MANIFEST.MF`**. The **`MANIFEST.MF`** files still have some minor differences, but the important information is the same. Please consider merging this PR in order to reduce the differences between the two generated jars and to provide the same info to Maven developers.

Thanks,
\- Kyle

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -44,6 +44,7 @@
 	<properties>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<project.reporting.outputEncoding>${project.build.sourceEncoding}</project.reporting.outputEncoding>
+		<maven.build.timestamp.format>yyyyMMdd-HH:mm:ss</maven.build.timestamp.format>
 
 		<!-- compiler plugin configuation -->
 		<maven.compiler.source>1.6</maven.compiler.source>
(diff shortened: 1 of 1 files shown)


zhedar:
Maybe we shouldn't add `BootsFaces OSP Gradle builder.`, if the artifact was not built with Gradle.
Otherwise I find this additions great, would have another look at this today before merging though.

stiemannkj1:
Oops! I changed it to  `BootsFaces OSP Gradle builder.` now, but I'm not sure if you want something else entirely.

zhedar:
Yeah including `Maven builder` is better, it may just help to distinguish the different build, that's all. :) (So the actual text is probably not that important.)

I think we can merge this as is now.

stephanrauh:
@stiemannkj1 Once again, thanks Kyle! It's a great honour that the LiferayFaces team starts to send us pull request!

Regards,
Stephan

stiemannkj1:
@stephanrauh :) You're welcome! Glad I could help!

stephanrauh:
@stiemannkj1 I don't find your original post, so I'll simply put my answer here. Today, you've mentioned that the LiferayFaces team didn't find the time to test BootsFaces with LiferayFaces yet. I'm looking forward to this test. But at the same time, I'm skeptical. As far as I remember, LiferayFaces doesn't work without Liferay, and both Liferay and Bootstrap are very opinionated frameworks. They bring their own CSS framework. Trying to support both BootsFaces and AlloyUI at the same time sounds like a real challenge to me. It's possible, but I don't think anybody will do it. 

Now it's up to you to prove I'm too pessimistic! :)

stephanrauh:
@stiemannkj1 Ah, that's interesting. I didn't know that. Do you still use AlloyUI? Or is AlloyUI compatible to Bootstrap?

stiemannkj1:
@stephanrauh, AlloyUI 3 is compatible with Bootstrap 3 and is used in Liferay 7.

PR #416, opened 2016-06-16
Datatableoptions

jottyfan:
Do not include this pull request yet, as this is not finished. Just for discussion and experiments, one may try these changes.


stephanrauh:
Congratulations, you've successfully submitted your first pull request to BootsFaces!

I understand from the discussions that your PR isn't finished yet? Just tell us when you're done.

stephanrauh:
I've set this PR to 0.9.1 because chances are we publish 0.9.0 this weekend, which may be a bit early for finishing the PR, integrating it and testing in meticulously. However, I'd love to include it in 0.9.0, so I don't rule out we're going to do so :).

jottyfan:
Of course do you define the version where this PR is included. Just to mention it, I've added maven dependencies for unit test, mockito and powermock. Also, I needed a json lib, so I added simple json from google. Maybe this is sth. to discuss before you accept the PR?

jottyfan:
Again, I'm stuck in ajax call generation - the last added commit contains a bootsfaces ajax call as I imagine it to be correct, but it seems to be damaged, as tests show that. I do not understand what's wrong and need a helping hand. Any ideas?

stephanrauh:
I've started to pull your PR into my workspace. At first glance, I observe that you seem to have missed the BootsFaces.jsfdsl file (just like everybody else does :) ). That's not a big deal, I just thought you might be interested. The constant in the Enum should be called exactly like the getter and the setter (apart the the prefix get- and set-, of course). If you use my Eclipse plugin (see https://github.com/stephanrauh/JSFLibraryGenerator/tree/master/plugins), this happens automatically. The plugin generates the taglib, the *Core class, parts of the documentation and even skeletons of the other classes involved.

stephanrauh:
Oops, seems I have to extend the plugin for your property class :).

stephanrauh:
Do I get it right: the property bean is optional, if it's not there, the data are stored in the component instead?

stephanrauh:
I considered your PR good enough to be merged. You can continue using your branch to send us another PR (at least I think so), but you'll want to merge it before. BTW, I'd like to add one or two changes to the datatable, so maybe it's a good idea to wait until I'm finished.

As for your AJAX problem: actually, I solved it by ignoring it. I suppose we can add an option to send every change of the datatable to the client immediately using an AJAX request, but I don't think many people need that feature.

Thanks for your good work!

BTW: Are you on Xing? I tried to look up your name, but your name doesn't seem to be a unique primary key :).

jottyfan:
First of all: big thanks for approving my pull request. I really feel honoured. Also thanks for your support and all positive criticism. I feel myself very well supported by you.

Nevertheless, this PR has not yet been ready to make it to the stable branch. The deeper I dig into your code, the more do I expect big changes. For example, I found out that there is still some storage for default page properties, but not yet used for storage. So I wonder if it is a good idea to add a new page storage, maybe it's better to adapt the default values for this...

Besides, without ajax support, that extension doesn't make any sense. Most data tables are in view mode only so that there's no request on leaving the view. At least a "save page options" button would be needed for non ajax usage to send the options to the bean.

On designing, I thought the new attribute propertyBean to be optional: if set, ajax requests are forced, if not, no ajax request is rendered. So if the tag is used, an implementation of the abstract bean must be bound to it. There's no plan to store the property bean in an abstract way.

I am still stuck on adding a simple ajax call on the hidden input field's onchange event. Would you mind to give me a hand  on this?

jottyfan:
Oh - and I'm not on Xing or any other job portal. Currently, I feel very comfortable with my employee, there's no need for other jobs.

jottyfan:
And another thing... are you aware of the extensions to pom.xml than adds mockito, junit and myfaces test libs to the project (all in test scope)? I also added google's json lib, maybe you want to remove so many new dependencies? You've still merged them...

stephanrauh:
I just saw this option: https://datatables.net/examples/basic_init/state_save.html

It seems this solves the problem much more simple and elegant. So I'd like to remove your code (which is a pity!) and replace it with a simple client-side option. Any objections?

stephanrauh:
> Nevertheless, this PR has not yet been ready to make it to the stable branch.

Sure... but I'd like to finish the data table soon, so I decided to finish the PR myself. You were definitely on the right track, and you've already solved most problems. However, it seems another guy has implemented the same feature in #279, in a very similar fashion. In any case, I think the `saveState` option of the widget itself does the job, too. We should add AJAX support for the pagination, sort etc. events.

As for Xing: I didn't want to hire you (although that might be nice, too :)). I just thought of extending my network.

jottyfan:
So you found a more elegant way to save the page state - this is what I wanted to have anyway. So I'll close my branch. Great to have this feature soon. Thanks for all your help and support!

PR #426, opened 2016-06-21
BUGFIX: non responsive button groups vertical alignment corrected

jottyfan:
When using buttongroup on a non responsive environment, rendering the styleClasses introduced a bug on vertical buttons. I removed the obsolete spaces and added a condition on responsive style class additions to prevent spaces between btn-group and its appendix -vertical.

Doing so, I also added some unit test libraries - you may decide if they are needed in the main branch.


zhedar:
Thanks for your contribution!

How did that bug look like, do you have an example / screenshot, so we can validate that?

Concerning the proposal: I personally don't really like `StringBuilder`, [because it's much less readable and the compiler probably optimizes that automatically](http://stackoverflow.com/questions/4645020/when-to-use-stringbuilder-in-java). But I've seen it being used quite frequently in the project, so there must be still someone who thinks, it's worth it. Personally, I think it isn't. @stephanrauh, @TheCoder4eu  and @asterd may think otherwise though ;)

jottyfan:
As requested, I've build a bug showcase here: https://github.com/jottyfan/MyfacesPlayground - just clone both branches (master for 0.9.0-SNAPSHOT and workaround for 0.9.0-DTO (included in https://github.com/jottyfan/BootsFaces-OSP )), deploy both on your tomcat and see the difference.
Also here's a screenshot that simply makes clear what happened:
bootsfaces_buttongroupbug

My proposal is just optional; I like StringBuilders and prefer them for all string concatenations because of their runtime speed (as you can read on the page from stackoverflow that you've linked). I know that the jvm is smart enough to convert that on compile time, but for me, the code feels more optimized with string builders, and I expect developers to be able to handle them. But you may reject this, the bug solution is one commit before that.

TheCoder4eu:
@jottyfan thanks for your contribution! :-)

TheCoder4eu:
@zhedar starting with Java 6/7 I had to change my mind on how good is Java at handling Strings.

I still need to have a look with a decompiler to what happens now (Java 8) with StringBuilder statements .

TheCoder4eu:
@jottyfan of course for large string processing a StringBuilder is faster (unless you use bad string handling defaults) for short Strings however you have to take in account the overhead of allocating a Big object like StringBuilder + the objects he needs.

jottyfan:
Ok, I can live with the string concatenation, never mind. But I think this bugfix should make it to the main branch before 0.9.0 is released.

stephanrauh:
@jottyfan Is this PR ready to be merged?

jottyfan:
Yes, it is ready for a merge, if you accept the StringBuilder changes. Otherwise, just merge only the first commit.

stephanrauh:
Ah, I see what's happened. A stupid bug, introduced by adding the same thing in 40 components, which made me sloppy after a while :).

stephanrauh:
As for StringBuilder: usually, I don't use it because I prefer my code to be as readable as possible. But that doesn't mean it's wrong to use it, especially when many string fragments are concatenated.

stephanrauh:
I've moved the responsive classes into a div of their own. Otherwise, they modify the width of the buttons, which may be the desired effect, but probably only in rare exceptions.

PR #451, opened 2016-07-06
Full calendar

jottyfan:
An extension to include http://fullcalendar.io/ as a component
```xml
<b:fullCalendar />
```
with basic properties. This component is missing value changes to go back to backing bean, but still showing dates is possible.
Current properties can be found in bootsfaces-b.taglib.xml, and for the events there can be used a helper classes toJson method in net.bootsfaces.component.fullCalendar.FullCalenderEventList.

Any suggestions are welcome, also improvement hints on the source code itself.


stephanrauh:
Last time I was a bit early to merge your PR, so this time I'll ask you beforehand: do you want us to inspect and merge the PR already?

jottyfan:
Well, this time it is not that dangerous because this is not a change on an existing component. Therefore, yes, if you may, have a look at the changes and find a decision if you want to accept this PR now or reject it because of lack of some expected functionality. Nevertheless, I have plans to extend the component by more attributes later when I find some time for it. But for now, the component works, and so I think this could be a starting point. In documentation and showcase, you could mark this component as a snapshot version - the way you did it with the datatable component. But in the end, that is your decision. Have a look, think about it and approve or reject this PR as you wish. I'd be glad to see this component in the next official release (0.9.1).

jottyfan:
After merging the 0.9.2-SNAPSHOT master into the current pull request, I think that the full calendar plugin can be introduced. There is still some work left, but with all the features that currently are available, one is able to use the calender, add javascript events on changes and bind jsf interactions on that.

stephanrauh:
I've started to integrate your PR. Currently, I've got two questions:

- Is there a documentation page? If not, I'll write it from scratch.
- Is there an updated BootsFaces.jsfdsl file?

stephanrauh:
I've integrated your component and started to document it. I hope I've got your default values right when I reverse-engineered the bootsfaces.jsfdsl file. 
http://www3.bootsfaces.net/Showcase/miscellaneous/FullCalendar.jsf

jottyfan:
Hi, sorry for the late answers, being busy all the time... But for now: very big thanks to include my PR! I'm glad to have this component available in one of the next releases. I didn't have a documentation page and also not a jsfdsl file, but you've still managed it all as I can see. Only one thing could be improved: on the showcase page, you could make the http url link become a real a-Tag so that people could go to that site directly - this is in the description field of the attribute "events". All other things are correct (due to the javadoc comments).

As I mentioned, extensions are supposed to improve the component, especially one to get all the events by clicking on a "make the changes persistent" button or such. But this is to be developed as soon as I can find some time. For such I will create new branches and pull requests.

Again, thanks for approving my suggestions!

stephanrauh:
Actually, it's up to us to say "thank you" :). Here you are: Thank you!

Currently, the showcase page is only a sketch. I'm going to improve it. And your right, there should be a way to make changes persistent. If you'd like to add this feature yourself, we'd appreciate this. If not, we'll take care of it, but it might take some time.

PR #503, opened 2016-09-11
DateTimePicker Missing Closing Div

MitchTalmadge:
The reason for this pull request is that in the DateTimePicker the closing div tag was incorrectly placed within a block of code that does not always run. Therefore, only popup type DateTimePickers would ever have the correct closing div. Also, the console was being spammed with the date format on every render of this component.

See this example of why the div change is necessary:

image

In this image, the dark blue highlighted div is the first DateTimePicker. The light blue highlighted div (within the first DateTimePicker) is the second DateTimePicker. The first DateTimePicker is engulfing the second, when they should be separate. This made it impossible to put the pickers into separate columns and caused many other minor formatting issues.


stephanrauh:
Thank you very much!

Issue #506, opened 2016-09-16
dateTimePicker minDate "Unexpected identifier"


MitchTalmadge:
I am trying to use `b:dateTimePicker` and having very little luck. Here is my code:

```xml
<b:dateTimePicker showDate="true" showTime="true" format="MM/DD/YYYY h:mm a"
                  minDate="#{timeSelectionController.minDateTime.toDate()}"
                  value="#{searchByTimeController.startTime}"/>
```

the minDateTime getter is simply: (Using joda time)

```java
public DateTime getMinDateTime() {
    return new DateTime();
}
```

The picker shows up:
![image](https://cloud.githubusercontent.com/assets/8848080/18576047/02fc8f36-7b9b-11e6-854e-6ea63594f364.png)

But clicking on it does nothing and I get this in the console on page load:
![image](https://cloud.githubusercontent.com/assets/8848080/18576060/1a1ae942-7b9b-11e6-9a31-39a32205808c.png)

Am I doing something wrong?

MitchTalmadge:
For the record, the value doesn't appear to be the problem -- setting it to null has the same effect.

asterd:
Hi @MitchTalmadge,

the problem is related to two main things:
1. A little bug inside the code that I've just fixed and committed to the repo.
2. Another little bug in your code (at least in my local test): if you set the minDate as the "toDate()" value, it returns a format that moment does not recognize (eg. 'Fri Sep 16 09:31:23 CEST 2016'). You must fix that.

I hope I've helped you.

MitchTalmadge:
Alright, here is my new code which prevents the error from being thrown, but does not perform the function I expect:

```xml
<b:dateTimePicker format="MM/DD/YYYY h:mm a"
minDate="'#{timeSelectionController.dateTimePickerMinDate}'"
value="#{searchByTimeController.startTime}"/>
```

Here is the minDate now:
```java
return new SimpleDateFormat("MM/dd/yyyy h:mm a").format(getMinDateTime().toDate());
```

(I added single quotes to minDate until the new code with your fix is released)

This works, except that the min date appears to have no effect at all on the picker. I can still select previous days and times on the calendar:

Actual calendar:
![image](https://cloud.githubusercontent.com/assets/8848080/18594955/ae931446-7bff-11e6-8bdc-8d7004b0e189.png)

Javascript: (notice the min date)
[...]

PR #507, opened 2016-09-17
Fixed DateTimePicker min/max date having no effect

MitchTalmadge:
https://github.com/TheCoder4eu/BootsFaces-OSP/issues/506

I found out that if the square brackets are removed, the min/max dates work as expected. I tested this and it does have the intended effect after the removal.

### DIFF
--- a/src/main/java/net/bootsfaces/component/dateTimePicker/DateTimePickerRenderer.java
+++ b/src/main/java/net/bootsfaces/component/dateTimePicker/DateTimePickerRenderer.java
@@ -46,11 +46,11 @@ public class DateTimePickerRenderer extends CoreRenderer {
 		DateTimePicker dtp = (DateTimePicker) component;
 		String clientId = dtp.getClientId();
 		String subVal = context.getExternalContext().getRequestParameterMap().get(dtp.getClientId());
-		
+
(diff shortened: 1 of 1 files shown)


MitchTalmadge:
I added more commits to fix the problem that the fields were not properly being disabled due to an incorrect javascript function.

Before:


After:
image

asterd:
Thank you very much @MitchTalmadge, I've merged your PR and as soon as I'll find some free time, I'll going to review this component to improve it or fix some other little bugs.

MitchTalmadge:
@asterd Thanks! I'd like to point out something I discovered this morning. By default, the way the datetimepicker library disables inline pickers...

image

...is to simply remove it completely. I found that a better way to do it is to give the outermost div these CSS properties:
```
opacity: 0.65;
pointer-events: none;
```

Which makes it look like this: (And nothing can be clicked)

image

So what I'd do is only add the javascript-based disabling to popup and basic type DateTimePickers, and let the inline type have those css properties instead. What do you think?

asterd:
Hi @MitchTalmadge, i like your solution! I don't know if this can have problems with ie, but if it does not, it's look like a real good solution from an ui point-of-view. So, can you please create a new PR with this trick?

Issue #482, opened 2016-08-02
Add AJAX to b:dateTimePicker


stephanrauh:
As @ScreamingTree86 points out, we've forgotten to add AJAX to `<b:dateTimePicker />`. BootsFaces 1.0 should fix that.

asterd:
Done and documented with 9e9b7e0 . Maybe @stephanrauh you must have a look to check is all ok, please!

stephanrauh:
I guess I should run a QA session on it, but it's definitely far better than before.

MitchTalmadge:
Using 0.9.2, I can't seem to get ajax to do anything at all. Was this tested?

asterd:
Hi @MitchTalmadge, the source code on GitHub was aligned with the support of Ajax in DateTimePicker (9e9b7e0 and 3532a41). I don't know if the pre-build jar contains this commit... but also in BootsfacesWeb (source code) the documentation is aligned with an example of ajax call.

MitchTalmadge:
@asterd It appears that ajax does work for popup and basic type fields, but not at all for inline. Also, the ajax appears to only fire when I click on the field, but not when I click on the dates inside the popup. For example, if I click on the field, which makes the popup appear, then ajax is fired. If I click a date inside the popup, even though the date changes inside the field, ajax is not fired. If I click away from the field, and then click on it again, ajax is now fired and the correct date is sent.

MitchTalmadge:
Found the problem; ajax is only happening on the `onclick` javascript event. To get it to work with the date change event, I just added `ondtchange="ajax:true"` to the dateTimePicker. I would have assumed that `ajax="true"` was all I needed for both onclick and ondtchange, so maybe that can be an improvement in the future?

TheCoder4eu:
@MitchTalmadge thanks for your PR!

TheCoder4eu:
@stephanrauh , I guess the PR solved the Issue? Or there was something left?

PR #508, opened 2016-09-17
Added support for DateTimePicker inline disabling; Fixed DateTimePicker defaultDate not working; Fixed inline DateTimePicker not setting value correctly.

MitchTalmadge:
The first part of the commit adds the functionality explained here: https://github.com/TheCoder4eu/BootsFaces-OSP/pull/507#issuecomment-247788085

The second part fixes the `initialDate` attribute of DateTimePicker (which equates to `defaultDate` in javascript) having no effect in the same way that `minDate` and `maxDate` did not.


MitchTalmadge:
Added another commit to solve the problem as stated here: https://github.com/TheCoder4eu/BootsFaces-OSP/issues/482

MitchTalmadge:
Added another commit because the `date` attribute of the date time picker library was only being set on inline dateTimePickers. It really should be set on all modes:

image

If this is not set, the date chosen may not match the value passed in, which is completely unexpected. Sometimes it's the minDate, sometimes something else, but it's completely unpredictable -- even when the value is not null.

MitchTalmadge:
Actually please hold on merging this for a bit, I need to test it a bit more.

MitchTalmadge:
Ok, all good to merge. 

The reason for the downgrade of the javascript file is because the latest version has a bug that causes the selected time to automatically revert back to whatever it was in the beginning when switching days. Not only that, but when using minDate, times would often be disabled for selection even when they were valid, which prevented the user from selecting the desired time.

asterd:
Hi @MitchTalmadge, thank you very much for your contribution: as I've already said, there is a lot of work to do, and sadly the time is ever the same! It would be gracefully to work more time on this project but is not always possible: so, all your contributions are well accepted, and every issue open let us improve our framework! Thank you!

MitchTalmadge:
@asterd No worries, friend! I'm here to help. The framework you guys have created has enabled me to carry out my dreams in less time than I had ever thought possible. This is the least I can do.

PR #509, opened 2016-09-19
Fixed navCommandLink using the wrong cursor. Fixes #479

### LINKED ISSUE #479: b:navCommandLink - different cursor when outcome & action is used
When I use b:navCommandLink's "outcome" attribute - there is a "hand" cursor.
When I use b:navCommandLink's "action" attribute - there is a "pipe" cursor. In "action" attribute there is a backend bean method called and navigation string is returned by this method.
 
I think that hand cursor should be shown in both cases.

I tested using version 0.9.1 and Firefox 47.

MitchTalmadge:
See #479 for more details. This simply adds a new class, `commandLink`, to the `<a ... />` part of the link if it is a navCommandLink. The class sets `cursor: pointer;`.


MitchTalmadge:
Yeah, that would be a good idea :)

Issue #501, opened 2016-09-08
navCommandLink works different: partialResponse


jottyfan:
Using navCommandLink that way:
````xml
<h:form>
  <b:navBar fluid="true">
    <b:navbarLinks pull="left">
      <b:navCommandLink active="#{myBean.active}" action="#{myBean.toPage}" value="page" />
    </b:navbarLinks>
  </b:navBar>
</h:form>
````
gives me that response:
````XML
<?xml version="1.0" encoding="UTF-8"?>
<partial-response id="j_id__v_0">
<changes>
<update id="javax.faces.ViewRoot">
<![CDATA[<!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/xhtml"><head>
...
````
Why do I not get a complete new page as response as when I use
````xml
<b:commandButton action="#{myBean.toPage}" value="page" />
````
? This is what I would expect...
The problem is, that after getting the partial response, all my b:panel objects do not toggle anymore, so sth. gets damaged on that partial response thing. Is there a way to make the navCommandLinks return a complete page instead of a partial response (didn't find an attribute for that now)?

MitchTalmadge:
I think if you want it to redirect you need to add `<redirect />` to your navigation case in `faces-config.xml`... see here: https://www.mkyong.com/jsf2/jsf-page-forward-vs-page-redirect/

I'm not entirely sure this is the problem but try it out at least

stephanrauh:
You could also try to set `ajax="false"`. If that doesn't work, I'd consider it a bug.

jottyfan:
@MitchTalmadge: no, I didn't want to redirect, I just wanted to get a complete html page as answer to my request as the commandButton does.

@stephanrauh: confirmed this behaviour as a bug. I added ajax="false" attribute to all my navCommandButtons:

````xml
<b:navBar fluid="true">
  <b:navbarLinks pull="left">
    <b:navCommandLink action="#{...}" active="#{...}" value="#{...}" ajax="false" />
    ...
  </b:navbarLinks>
</b:navBar>
````

but the answer is still a partial response like that:

````xml
<?xml version="1.0" encoding="UTF-8"?>
<partial-response id="j_id__v_0">
<changes>
<update id="javax.faces.ViewRoot">
<![CDATA[<!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/xhtml"><head><meta name="viewport" content="width=device-width, initial-scale=1"/>
````

[...]

PR #528, opened 2016-10-20
do not render ajax when disabled

jottyfan:


jottyfan:
There is an issue (#501) that is about the change request.

stephanrauh:
Seems you've found the bug pretty soon. Thanks!

jottyfan:
Even with less time, I tried to set up a unit test environment for the navlink renderer class. This test is not yet finished, but if you may have a look at it, you could find the reason for the damaged html code maybe. I've copied and modified some classes from tutorials and stack overflow solutions to make the unit test run.

https://github.com/jottyfan/BootsFaces-Tests/blob/mockingtests/src/test/java/net/bootsfaces/component/navLink/TestNavLinkRenderer.java

If you run that test class, you will end up in a shell output without valid html code. Maybe there are some attributes missing, but I'd like to find out what's wrong with the renderer on using ajax=false. Any ideas?

BTW, BootsFaces-Tests looks like orphaned - is this really the test framework for bootsfaces? I'd expect much more test classes (at least one for a component renderer).

PR #589, opened 2016-12-18
Fixes datetimepicker update by ajax call before completly initialized

### LINKED ISSUE #586: DateTimePicker fails with ajax update
DateTimePicker (Github master) destroys date value during ajax call immediately after rendering the page.

I've initialized a date on initialization for two date time picker controls. During the first rendering, before touching the picker with an user action. DateTimePicker sends back a date to the server with a wrong year. 

I have attached a small example showing this issue. I've tried to debug, rendering looks good. Seems to be the Ajax call, which may be to early, with an not fully initialized date value.

**Some output from the DateTimePickerRenderer:**

DTP:Thu Sep 01 19:41:26 CEST 2016
DTP sloc:de
DTP format:MM.YYYY
DTP displayformat:'MM.YYYY'
DTP inlineDisplayDate:'09.2016'
DTP:Fri Mar 31 19:41:26 CEST 2017
DTP sloc:de
DTP format:DD.MM.YYYY
DTP displayformat:'DD.MM.YYYY'
DTP inlineDisplayDate:'31.03.2017'

**Output from DTPBean:**

set1:Wed Mar 09 00:00:00 CET 7
set2:Wed Mar 09 00:00:00 CET 7

**Example files:**

[DTPBean.java.txt](https://github.com/TheCoder4eu/BootsFaces-OSP/files/655546/DTPBean.java.txt)
[dtptest.xhtml.txt](https://github.com/TheCoder4eu/BootsFaces-OSP/files/655547/dtptest.xhtml.txt)

mtvweb:
Using a b:datetimepicker with ajax update was sending an invalid (ISO)
format date to the server before initialization of the datetimepicker
itself. Moving event initialization to the jquery ready handler fixes
this behaviour.

Fix #586

### DIFF
--- a/src/main/java/net/bootsfaces/component/ajax/AJAXRenderer.java
+++ b/src/main/java/net/bootsfaces/component/ajax/AJAXRenderer.java
@@ -632,6 +632,22 @@ public class AJAXRenderer extends CoreRenderer {
 	 */
 	public void generateBootsFacesAJAXAndJavaScriptForJQuery(FacesContext context, UIComponent component,
 			ResponseWriter rw, String jQueryExpressionOfTargetElement, Map<String, String> additionalEventHandlers) throws IOException {
+		generateBootsFacesAJAXAndJavaScriptForJQuery(context, component, rw, jQueryExpressionOfTargetElement, additionalEventHandlers, false);
+	}
+	
+	/**
+	 * Registers a callback with jQuery.
+	 *
+	 * @param context
+	 * @param component
+	 * @param rw
+	 * @param jQueryExpressionOfTargetElement
+	 * @param additionalEventHandlers
+	 * @param attachOnReady
+	 * @throws IOException
+	 */
+	public void generateBootsFacesAJAXAndJavaScriptForJQuery(FacesContext context, UIComponent component,
+			ResponseWriter rw, String jQueryExpressionOfTargetElement, Map<String, String> additionalEventHandlers, boolean attachOnReady) throws IOException {
 		if (jQueryExpressionOfTargetElement.contains(":")) {
 			if (!jQueryExpressionOfTargetElement.contains("\\\\:")) { // avoid escaping twice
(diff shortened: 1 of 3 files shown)


stephanrauh:
I've tried to integrate your bug fix. At first glance, it looks great, but there seem to be at least two issues:

* I had to simplify your reproducer because I don't have your template file. After that, jQuery wasn't available, resulting in errors like "Uncaught ReferenceError: $ is not defined".
* I wonder if your `$(function() { ... }` (which is equivalent to `$( document ).ready(function() { ... });` works if the datePicker is created by an AJAX request. The difference being that `ready()` probably is never called. But I didn't test it yet, so it's just guesswork.

Nonetheless: thank you very much for providing the pull request and the great reproducer!

Best regards,
Stephan

mtvweb:
Sorry, I had only stripped down a part of my project to the relevant code, but forget about the layout.  I've changed it now to a standalone one. Also I added some more to test the update case. Which is working now well for me.

- Regarding the ReferenceError. Because the datetimepicker is using the same $(function() ....) some lines before for initialization in DateTimePickerRenderer, so it should be there too. Maybe something was missing and the updated test files help.

- The initialization should work, even with ajax inserted in the DOM. My experience (by JSF and non JSF projects) is that JQuery schedules them quite well, even if the document was already initialized. The advantage is that they are executed in the same sequence as they are added to the ready handler. You are using $() initialization for DateTimePicker initialization which was working well too for ajax requested ones.

Regards
Jens 

[DTPBean.java.txt](https://github.com/TheCoder4eu/BootsFaces-OSP/files/660557/DTPBean.java.txt)
[dtptest.xhtml.txt](https://github.com/TheCoder4eu/BootsFaces-OSP/files/660558/dtptest.xhtml.txt)

stephanrauh:
@mtvweb Sounds good! I'm sorry that your PR didn't make it into the 1.0.1 version. We thought about integrating it, but it would have really been a last minute change, so we decided to err on the safe side. There simply was no time to test it. I hope I managed to run and integrate the tests later this week.

PR #600, opened 2017-01-01
b:datetimepicker issue after changing id / b:radioButton ajax update

mtvweb:
Find another fix for dateTimePicker which was not working well after last "id" update, which switches from clientId to fieldId. clientId was still used for JQuery. 

Also it contains an improved b:radioButton ajax update handling (maybe the code in Ajax decode could be better, but for now it works) updating bean value.

See:
https://github.com/mtvweb/BootsFaces-OSP/commit/ba9bb263065c4d0cbf0cf6ca2fc5c4589af22ba7
https://github.com/mtvweb/BootsFaces-OSP/commit/ecce9cb28191830d420f6081d64707b47ba5f503


stephanrauh:
Thanks! BTW, when you change the components, please also update the BootsFaces.jsfdsl. This allows your change to survive a bit longer. We use the XText plugin (https://github.com/stephanrauh/JSFLibraryGenerator) to generate a lot of code from the DSL file.

Issue #591, opened 2016-12-18
Ordering a b:dataTableColumn by date


ruanfelisky:
I'm trying to sort a `<b:dataTableColumn>` whose data type is a date (dd/MM/yyyy), but it seems like it's not doing properly, even setting the property `dataType="date"`. I've also tried with `orderBy="dom-text"` or `orderBy="dom-text-numeric"` but no success.

Please see the example below:
```java
@NotNull
@Temporal(TemporalType.DATE)
@Column(name = "data_movimento")	
public Date getDataMovimento() {
	return dataMovimento;
}
```
```xml
<b:dataTableColumn label="Data" dataType="date">
     <h:outputText value="#{movimento.dataMovimento}">
          <f:convertDateTime pattern="dd/MM/yyyy" />
     </h:outputText>	    				
</b:dataTableColumn>
```
![image](https://cloud.githubusercontent.com/assets/4307334/21294502/5289a7a4-c525-11e6-9cbe-1a2608b66695.png)

Issue #369, opened 2016-05-07
Publish Developer Snapshots on MavenCentral


stephanrauh:
Every once in a while, I upload an untested developer snapshot of BootsFaces to the Sonatype repository, so you can test bug fixes and new features early.

**Important: Do not use these snapshots in production!**

https://oss.sonatype.org/content/repositories/snapshots/net/bootsfaces/bootsfaces/0.9.0-SNAPSHOT/

You can access this in your pom.xml after adding a new repository:
``` xml
     <repositories>
        <repository>
          <id>central</id>
          <url>https://oss.sonatype.org/content/repositories/snapshots</url>
          <releases><enabled>true</enabled></releases>
          <snapshots><enabled>true</enabled></snapshots>
        </repository>
    </repositories>
```

The download coordinates of the 0.9.0 snapshot version are:
``` xml
<dependency>
	<groupId>net.bootsfaces</groupId>
	<artifactId>bootsfaces</artifactId>
	<version>0.9.0-SNAPSHOT</version>
	<scope>compile</scope>
</dependency>
```

stephanrauh:
Just pushed a new version of BootsFaces 0.9.0-SNAPSHOT on Maven Central.

stephanrauh:
Just uploaded another snapshot of BootsFaces 0.9.0. It contains a lot of improvements to the dataTable.

stephanrauh:
Just uploaded a 0.9.1-SNAPSHOT to Maven Central. Basically, it's a hotfix addressing the Java 6 and 7 incompatibility. More details at https://github.com/TheCoder4eu/BootsFaces-OSP/milestone/14?closed=1

mannan4net:
Hi Team, Tried to use the snapshot for one of the issue for datatable resolved but facing issues.

Failed to execute goal on project MavenConfTest: Could not resolve dependencies for project com.mycompany:MavenConfTest:war:1.0-SNAPSHOT: Failure to find net.bootsfaces:bootsfaces:jar:1.0.0-SNAPSHOT in https://oss.sonatype.org/content/repositories/snapshots was cached in the local repository, resolution will not be reattempted until the update interval of central has elapsed or updates are forced ->

Please help to get the snap shot

stephanrauh:
Sorry, the fault is on my side. For some reason, I confused the version numbers. The correct version number of the snapshot is 0.9.2.

stephanrauh:
[...]

PR #602, opened 2017-01-02
b:dataTableColumn added properties for data-order and data-search

mtvweb:
Allows to specify values for ordering or searching. Useful i.E. with
formatted values (like a localized Date).

Example:
<b:dataTableColumn label="Date" order="asc" data-order="#{
dateValue.toGMTString() }" data-type="date">
	<h:outputText value="#{ dateValue }">
		<f:convertDateTime dateStyle="medium" />
	</h:outputText>
</b:dataTableColumn>

Fix TheCoder4eu/BootsFaces-OSP#591


stephanrauh:
Once again: thanks!

stephanrauh:
I've published a new developer snapshot containing your PR on Maven Central. As usual, see #369 on how to get it.

chongma:
using 1.0.2 if i pass a date object to data-order it doesn't work.  do i need to convert it to a string?  shouldn't the default behaviour be that it orders by date?  i.e. yyyy-MM-dd HH:mm:ss

PR #620, opened 2017-01-18
added 'info' property to b:dataTable

mtvweb:
info property should be independent from paginated option. Offering more
flexibility.

I require a separation of the info display from the paginated option. Added option "info" instead hiding the info if "paginated" is disabled. Hope you like it.


stephanrauh:
Thanks! BTW: if you'd like to get in contact with us, drop a mail at webmaster1 at beyondjava.de.

Issue #619, opened 2017-01-17
b:dataTable not styled by patternfly


NicolaIsotta:
I was trying the patternfly theme and I like its "professional" look; the `b:dataTable` component, though, is not styled [like it should](http://www.patternfly.org/pattern-library/content-views/table-view/#_).
I think this is caused by:

- file `datatables.min.css` which should not be included when patternfly is the chosen theme (a context-param to control its loading would be useful too)
- the `core.css` file for patternfly theme should include tables and datatables customization
As always, thanks for your work.

NicolaIsotta:
I've just created a PR. This is how a b:dataTable appears now:
![datatable_patternfly](https://cloud.githubusercontent.com/assets/18428155/22431930/8b637914-e713-11e6-8ce8-f3e71fea8022.jpg)

TheCoder4eu:
@NicolaIsotta your PR looks ok to me, thank you very much for this contribution!

NicolaIsotta:
@TheCoder4eu unfortunately, the datatables customization by patternfly doesn't include the responsive plugin ([and they have no plans to do it](https://patternfly.atlassian.net/browse/PTNFLY-1967)), as you can see in the snapshot showcase. I've already prepared a PR which fixes the problem.

PR #632, opened 2017-02-04
#619 Improve datatables & patternfly integration

NicolaIsotta:


stephanrauh:
I'm not our expert for Patternfly, so I hesitate to merge your PR myself. It's better to wait a few days. Nonetheless, I'd like to say "thank you" for providing the PR!

stephanrauh:
@NicolaIsotta I'm sorry about the long delay. It's really embarrassing. It's high time to merge it!

Today, I've tried to understand your PR. Truth to tell, I didn't understand it. Would you mind to explain what you did? Of course, I could simply merge and test the PR, but I prefer to understand it :).

Questions to ponder:
* You've removed the code removing the include of `datatables.min.css`. Why don't we need it anymore? (Or was it always wrong, and you've fixed it?)
* You've removed the imports of `table-view.less` and `datatables.less`. Do these files generate the datatables.min.css file?
* Did you test the non-patternfly themes? I don't understand how these themes work without the `table-view.less` and the `datatables.less`.

NicolaIsotta:
Ok, let's try to explain this. Hope I'm remembering everything.
Beform my PR #630 , b:dataTable wasn't styled by Patternfly. So I enabled both tables and datatables customization in the bs-core.less file of patternfly. Also, the datatables.min.css wasn't needed when the theme was patternfly, and that was the reason for the first edit of the AddResourceListener.

Unfortunately, after visitng the snapshot showcase with the PR merged, I discovered that patternfly doesn't support the responsive plugin of datatables ([and they have no plans to support it](https://patternfly.atlassian.net/browse/PTNFLY-1967)). You can check by going [here](http://www3.bootsfaces.net/Showcase/forms/DataTable.jsf#responsiveDesign) and setting the theme to patternfly.

So, here's what this pr does:
- removes datatables customization from the patternfly bs-core.less, but leaves the the table customization
- comments out the the code which removed datatables.min.css when the theme was patternfly
- adds some css which improves datatables.min.css integration with patternfly-styled table. Basically, it involves the positioning of the sorting icons.

The final result should be like this:
table

stephanrauh:
@NicolaIsotta Thanks for the explanation! I think now I've got it. However, I'd like it to our team lead to merge the PR because most of the changes require our Gradle build. Problem is, my project setup is optimized for Maven. So it always takes me a while to get to speed with the Gradle scripts. Stay tuned, and bear with us!

Best regards,
Stephan

TheCoder4eu:
@stephanrauh if you agree we can merge.

Issue #637, opened 2017-02-16
Improvement: poll interval attribute type could be more lenient


jepsar:
Now it requires you to use a String. When you use EL to calculate a value on the fly, like `#{1000 * user.refresh}`, it will result in a Long which is not accepted. It would be nice if it accepts instances of String or Number (int value).

I'm on 1.0.2

jepsar:
I'll submit a pull request

stephanrauh:
I've merged the pull request. Todos left:

- [ ] testing
- [ ] updating the documentation

PR #638, opened 2017-02-16
Fix for issue #637

jepsar:
Sorry for the noise caused by trim trailing spaces on save...

### DIFF
--- a/src/main/java/net/bootsfaces/component/poll/Poll.java
+++ b/src/main/java/net/bootsfaces/component/poll/Poll.java
@@ -1,8 +1,8 @@
 /**
  *  Copyright 2015-2016 Stephan Rauh (http://www.beyondjava.net)
- *  
+ *
  *  This file is part of BootsFaces.
- *  
+ *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
 * You may obtain a copy of the License at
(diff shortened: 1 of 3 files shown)


stephanrauh:
At first glance, your PR looks good. Thanks! I don't have enough time to test it thoroughly, so I'll merge it this week-end.

Would you mind to updated the BootsFaces.jsfdsl, too? This is the file we use to generate the component classes, the taglib and much of the documentation.

Thanks in advance
Stephan

jepsar:
I've seen BootsFaces.jsfdsl, but don't know what to change there. It's not mentioned that interval should be a string, so it looks alright as is to me.

stephanrauh:
Well, yes, you're probably right. The default data type is String. But it's hard to get it right without installing the [eclipse plugin](https://github.com/stephanrauh/JSFLibraryGenerator). I'll care about it.

jepsar:
I was thinking that it might be useful to have `toInteger(Object object, int defaultValue)` in the utils class. Since when reading attributes in most cases (or even all) you want to have a default value. If you agree I will add this to my fork.

stephanrauh:
JSF already provides optional default values for component attributes, including integers and booleans. If you see a use case that's not covered by this, feel free to go ahead.

```xml
public class Column extends UIOutput ... {
   ...
   public int getTooltipDelayShow() {
     return (int) (Integer) getStateHelper().eval(PropertyKeys.tooltipDelayShow, 
                                                                            0 // default value
                                                                            );
   }
}
```

jepsar:
OK, clear. The poll component is not written this way (it gets the attribute values from a map). I'll leave it as is.

stephanrauh:
Thanks!

stephanrauh:
@jepsar May I mention you in "Kudos" section of the BootsFaces 1.1.0 announcement post?

jepsar:
@stephanrauh Sure. Glad I could help out.

Issue #652, opened 2017-03-08
b:commandLink


chongma:
i looked at b:button and b:commandButton and neither have a target or href attribute as far as i could see.  this is useful if you want to open something in a new tab.  maybe this could be a superclass of commandLink?
btw where do i register interest in helping with the project?  i have a couple of other things to mention about the documentation which i might be able to help with
- it would be nice if it was ordered alphabetically as it is hard to scan for a 
- if there were links to the source code of the examples directly from the page, or maybe a snippet of the backing bean it would sometimes help to understand the use case

stephanrauh:
`<b:commandButton>` doesn't need either attribute because it's meant to call JSF actions. However, it might be a good idea to add these attributes to `<b:button>`.  Actually, I'm surprised we didn't implement it yet.

As for contributing: simply drop us a mail at "webmaster1 at beyondjava dot de". Like you, I think there's a lot to improve with our showcase :). Instead of sorting stuff alphabetically I'd prefer to offer a search field. Same idea, just differently implemented.

stephanrauh:
Thanks for providing a pull request!

PR #655, opened 2017-03-14
href and target on b:button

chongma:
#652 
if the button is rendered as an anchor rather than a button in ResponseWriter it works okay and doesn't seem to affect anything.  does it specifically need to be a button?  outcome is called on target in onclick with window.open.  will this break backwards compatibility?


stephanrauh:
IMHO the button should be rendered as a button by default. This improved screen reader support and prevents compatibility problems. Most likely there's a project out there using CSS rules for our button component. So I've added some clever (or so I hope) logic to render either an anchor or an button. I've also added some documentation to the showcase.

That said: thank you very much!

chongma:
should it throw an exception if both outcome and href are set?

stephanrauh:
Good idea!

Issue #654, opened 2017-03-12
b:dateTimePicker format


chongma:
i just tried a b:dateTimePicker with value= java.util.Date and format="dd/MM/yyyy" and i get We/03/yyyy and other strange things in the picker.  with no format i get a short date.  am I using the wrong date format or passing the wrong object?

chongma:
it works with "DD/MM/YYYY".  is there a reason it doesn't follow the normal formatting that would usually be passed to SimpleDateFormat or f:dateTimeConverter for example?

stephanrauh:
Oh. It's good that you mention this. Actually, `b:dateTimePicker` is this component: https://eonasdan.github.io/bootstrap-datetimepicker/Options/, which is using the formatting options of moment.js. These options are completely different from what we Java programmers are used to: http://momentjs.com/docs/#/displaying/format/

Fancy to add some lines to the documentation page?

chongma:
Moved to BootsFacesWeb project [pull request](https://github.com/TheCoder4eu/BootsFacesWeb/pull/23)

stephanrauh:
Oops - it seems we've got a big problem. The formatting string "ddd, hA" of moment.js does not work. The method `DateTimePickerRenderer::getInternalDateAsString()` uses the `SimpleDateformatter`, passing the format strings of moment.js to it. This obviously only works with the small subset of common formatting options.

chongma:
I closed because I thought it was only a documentation issue. It should only use moment.js or the simpledateformatter but not both? I can check that function tomorrow to see if I can be of any help with the issue?

stephanrauh:
Yeah, it was a surprise to me, too. Honestly, I don't know how to solve the issue. I think the Java side simply expects a `java.util.Date`. Is it possible to display the date in the user-defined format, while passing it as a standardized format in the requests and responses? Such as "dd.mm.yyyy" or the number of milliseconds since 1970?

chongma:
could it use this? https://github.com/MadMG/moment-jdateformatparser
the user could define the format using Java notation and use this library to convert it to moment js format in the client at initialisation?  is there already a javascript block to initialise the control?

chongma:
actually...i mean the opposite to not break backwards compatibility.  the user defines using moment.js format.  it gets converted internally to Java format (for the SimpleDateFormatter) then when the control is rendered the Java format gets converted back to moment.js format.  Hope that makes sense!  Let me know and I will create a PR

[...]

PR #656, opened 2017-03-15
test singleton showing moment js formatting

chongma:
#654 
here is an example showing the functions in a singleton.  it is a nasty javascript eval hack but it seems to function ok.  the idea would be to initialise empty dateTimePicker with moment.js format and keep format in moment.js. and never convert to java format internally.


stephanrauh:
This PR seems to be compatible all the way down to Java 6, so I've merged it. However, I'm a bit concerned about performance. What about caching? And could we identify the most common use cases (like American date format and continental European date format) in order to hard-wire them?

Issue #682, opened 2017-04-06
Inputs with col-XX generate invalid Bootstrap markup


ggam:
With the following XHTML:

`<b:inputText value="test" col-xs="6" label="Message" label-col-xs="6"/>`

I'd expect it to generate something like:

```xml
<div class="form-group">
    <label for="message1" class="col-xs-6">Message</label>
    <div class="col-xs-6">
        <input type="text" class="form-control" id="message1"></input>
    </div>
</div>
```

But it's generating the following instead:
```xml
<div id="j_idt6:j_idt7" class="form-group ">
    <label for="input_j_idt6:j_idt7" class=" col-xs-6 control-label">Message</label>
    <input id="input_j_idt6:j_idt7" name="input_j_idt6:j_idt7" class="form-control" value="test" type="text"></div>
```

You can see that only the label has `col-xs-X` defined.

If I only set the size for the input, like `<b:inputText value="test" col-xs="6" label="Message" />`, I get the following:

```xml
<div class="col-xs-6" id="j_idt6:j_idt7">
    <div class="form-group ">
        <label for="input_j_idt6:j_idt7" class=" control-label">Message</label>
        <input id="input_j_idt6:j_idt7" name="input_j_idt6:j_idt7" class="form-control" value="test" type="text">
    </div>
</div>
```

`form-group` acts like `row` for forms, so it doesn't seem to be correct to wrap the row with a column. Also, the input size shouldn't be ignored when setting the label width. This is needed in order to allow for more complex form designs (such as https://jsbin.com/jikiyuwuve/edit?html,output). Although more customization options will be needed for that. I'll open an issue for that later.

stephanrauh:
The markup is wrong because you didn't add `<form horizontal="true">`. (Or at least I hope so because it was so hard and painful to get it right :)).

Have a look at http://www3.bootsfaces.net/Showcase/forms/horizontalForms.jsf for more details.

Should we emit an error message, or should we silently correct the html code the way you suggest? In that case, I'd like to postpone it to BootsFaces 1.2 because it's a breaking change and because it's very difficult to get it right. There are so many options (AJAX vs. non-AJAX, facets, horizontal forms, stacked forms, inline forms, and last not least input fields without the `col-*` attributes).

ggam:
Hi,

You are right, he form class is effectively needed for added styles. But
anyway, there are two problems with the markup not complaining with BS
rules (at least to my understanding):
[...]

Issue #614, opened 2017-01-09
Horizontal form only works in the most simple case


stephanrauh:
The other day I've learned that every component that is to be aligned vertically must be included in the some `form-group`. That's hard to accomplish with JSF because each component is rendered individually. We need to find a solution nonetheless.

One important special case is adding the `b:message` tag into the same line as the label and the input field. 

Other use cases are:

- Putting a command button behind an input field.
- Putting input fields behind a checkbox or a radio button.

NicolaIsotta:
#493 is related to this. As an alternative to the `facet`, have you considered modyfing the `encodeChidren` method?

stephanrauh:
I suppose this issue is solved by #682 and #685.

stephanrauh:
This issue has been solved by introducing `<b:form />`.

Issue #493, opened 2016-08-24
Add "after" facet to inputs


NicolaIsotta:
It would be useful to have an "after" facet in all the components who generate a form-group. The contents of this facet (for example, a `<b:message/>` or an `help-block`) should be rendered inside the form-group, but after the input/select/input-group.

stephanrauh:
Would you mind to show us an example demonstrating your use case? We can add the facet, but the documentation looks nicer if we can explain why we added it :).

stephanrauh:
related to #511.

NicolaIsotta:
Actually, in my case, this is related to horizontal forms (#485) . I'd like to achieve a layout where error messages and/or description test can be shown next to an input.
Just an easy example taken from web:
http://i.stack.imgur.com/ToPRZ.png

stephanrauh:
related to #614. In fact, #614 is the more general solution.

stephanrauh:
Solved by #682 and #685.

PR #685, opened 2017-04-08
New formGroup component

ggam:
As promised, here is the formGroup component. I extended it from Row as they are basically the same thing. 

I've also updated the InputText renderer to comply with it. This is just a prototype so you @TheCoder4eu @stephanrauh can tell me if I'm going in the right direction.

I have a limited connection now and I can't download XText so I copied the jsfdsl from the row component for now.

Could you please give me any advise? If this is ok, I'll change the rest of the input renderers, although that probably should be added before the div switch we talked about in #682.

### DIFF
--- a/src/main/java/net/bootsfaces/component/colorPicker/ColorPicker.java
+++ b/src/main/java/net/bootsfaces/component/colorPicker/ColorPicker.java
@@ -18,6 +18,7 @@ import net.bootsfaces.render.IHasTooltip;
 import net.bootsfaces.render.IResponsive;
 import net.bootsfaces.render.IResponsiveLabel;
 import net.bootsfaces.render.Tooltip;
+import net.bootsfaces.utils.BsfUtils;
 
 @FacesComponent("net.bootsfaces.component.colorPicker.ColorPicker")
 public class ColorPicker extends ColorPickerCore implements IHasTooltip, IAJAXComponent, IResponsive, IResponsiveLabel {
@@ -49,8 +50,7 @@ public class ColorPicker extends ColorPickerCore implements IHasTooltip, IAJAXCo
 		AddResourcesListener.addThemedCSSResource("core.css");
 		AddResourcesListener.addResourceToHeadButAfterJQuery(C.BSF_LIBRARY, "js/jquery.minicolors.min.js");
 
-		renderLabel = FacesContext.getCurrentInstance().getExternalContext()
-				.getInitParameter("net.bootsfaces.defaults.renderLabel");
+		renderLabel = BsfUtils.getInitParam("net.bootsfaces.defaults.renderLabel");
 		if (null != renderLabel && renderLabel.contains("#{")) {
 			renderLabel = ELTools.evalAsString(renderLabel);
 		}
--- a/src/main/java/net/bootsfaces/component/colorPicker/ColorPickerRenderer.java
+++ b/src/main/java/net/bootsfaces/component/colorPicker/ColorPickerRenderer.java
@@ -10,7 +10,7 @@ import javax.faces.render.FacesRenderer;
 
 import net.bootsfaces.C;
 import net.bootsfaces.component.ajax.AJAXRenderer;
-import net.bootsfaces.render.CoreRenderer;
+import net.bootsfaces.render.CoreInputRenderer;
 import net.bootsfaces.render.H;
 import net.bootsfaces.render.R;
 import net.bootsfaces.render.Responsive;
@@ -18,7 +18,7 @@ import net.bootsfaces.render.Tooltip;
 import net.bootsfaces.utils.BsfUtils;
 
 @FacesRenderer(componentFamily = C.BSFCOMPONENT, rendererType = "net.bootsfaces.component.colorPicker.ColorPicker")
-public class ColorPickerRenderer extends CoreRenderer {
+public class ColorPickerRenderer extends CoreInputRenderer {
 	private static final Logger LOGGER = Logger.getLogger(ColorPickerRenderer.class.getName());
 
 	@Override
@@ -102,7 +102,9 @@ public class ColorPickerRenderer extends CoreRenderer {
 			rw.writeAttribute("class", getFormGroupWithFeedback("form-inline", clientId), "class");
 			LOGGER.warning("The inline attribute of b:inputText is deprecated and generates faulty HTML code. Please use <b:form inline=\"true\"> instead.");
 		} else {
+                    if(hasToRenderFormGroup(component)) {
 			rw.writeAttribute("class", getFormGroupWithFeedback("form-group", clientId), "class");
+                    }
 		}
 		LOGGER.warning("The inline attribute of b:inputText is deprecated and generates faulty HTML code. Please use <b:form inline=\"true\"> instead.");
 
--- a/src/main/java/net/bootsfaces/component/dateTimePicker/DateTimePickerRenderer.java
+++ b/src/main/java/net/bootsfaces/component/dateTimePicker/DateTimePickerRenderer.java
@@ -32,14 +32,14 @@ import javax.faces.render.FacesRenderer;
 
 import net.bootsfaces.component.ajax.AJAXRenderer;
 import net.bootsfaces.component.icon.IconRenderer;
-import net.bootsfaces.render.CoreRenderer;
+import net.bootsfaces.render.CoreInputRenderer;
 import net.bootsfaces.render.Responsive;
 import net.bootsfaces.render.Tooltip;
 import net.bootsfaces.utils.BsfUtils;
 
 /** This class generates the HTML code of &lt;b:dateTimePicker /&gt;. */
 @FacesRenderer(componentFamily = "net.bootsfaces.component", rendererType = "net.bootsfaces.component.dateTimePicker.DateTimePicker")
-public class DateTimePickerRenderer extends CoreRenderer {
+public class DateTimePickerRenderer extends CoreInputRenderer {
 	private static final String DTP_OUTER_CONTAINER_SUFFIX = "Outer";
 
 	@Override
@@ -170,7 +170,9 @@ public class DateTimePickerRenderer extends CoreRenderer {
 			if (!isHorizontalForm(dtp)) {
 				rw.writeAttribute("class", responsiveStyleClass + " form-group " + errorSeverityClass, "class");
 			} else {
+                            if(hasToRenderFormGroup(dtp)) {
 				rw.writeAttribute("class", "form-group " + errorSeverityClass, "class");
+                            }
 			}
 			rw.writeAttribute("id", clientId, null);
 			Tooltip.generateTooltip(fc, dtp, rw);
--- a/src/main/java/net/bootsfaces/component/formGroup/FormGroup.java
+++ b/src/main/java/net/bootsfaces/component/formGroup/FormGroup.java
@@ -1,0 +1,28 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
+ * To change this template file, choose Tools | Templates
+ * and open the template in the editor.
+ */
+package net.bootsfaces.component.formGroup;
+
+import javax.faces.component.FacesComponent;
+import net.bootsfaces.C;
+import net.bootsfaces.component.row.Row;
+
+/**
+ *
+ * @author Guillermo González de Agüero
+ */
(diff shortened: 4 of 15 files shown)


stephanrauh:
At first glance, the changes look good. I'd just propose a change to the BootsFaces.jsfdsl file. Since the class is derived from `Row`, every attribute should be declared `inherited`. Otherwise, the attributes are generated although they are already in the parent class. Unfortunately, the "+" parameters can't be inherited yet - maybe I should update the Xtext generator again to support that use case.

```
widget formGroup
       implemented_by net.bootsfaces.component.formGroup.FormGroup
       has_children
       has_tooltip
 {
    dir                 inherited                                          "Direction indication for text that does not inherit directionality. Legal values: ltr (Default. Left-to-right text direction), rtl (Right-to-left text direction) and auto (let the browser figure out the direction of your alphabet, based on the page content)."
    id                  inherited                                          "Unique identifier of the component in a namingContainer."
    rendered            Boolean inherited                                  "Boolean value to specify the rendering of the component, when set to false the component will not be rendered."
    style               inherited                                          "Inline style of the input element."
    style-class         inherited                                          "Style class of this element."
    +tooltip
	+contentDisabled
}```

**Comment by stephanrauh** [comment]
The generated `FormGroup` class looks a bit different. Please add a few lines to the constructor:

```java
	public FormGroup() {
		Tooltip.addResourceFiles();
		AddResourcesListener.addThemedCSSResource("core.css");
		AddResourcesListener.addThemedCSSResource("bsf.css");
		setRendererType(DEFAULT_RENDERER);
	}
```

stephanrauh:
You'll also need the taglib (bootsfaces-b.taglib.xml). The generator adds these lines:
```xml
	
	<!-- *********** b:formGroup ************************* -->
	<tag>
		<tag-name>formGroup</tag-name>
		<component>
			<component-type>net.bootsfaces.component.formGroup.FormGroup</component-type>
		</component>
		<attribute>
		  <description><![CDATA[Enables or disables every child element of this container. By default, child elements are enabled.]]></description>
		  <name>content-disabled</name>
		  <required>false</required>
		  <type>java.lang.Boolean</type>
		</attribute>
		<attribute>
		  <description><![CDATA[Enables or disables every child element of this container. By default, child elements are enabled.]]></description>
		  <name>contentDisabled</name>
		  <required>false</required>
		  <type>java.lang.Boolean</type>
		</attribute>
		<attribute>
		  <description><![CDATA[Direction indication for text that does not inherit directionality. Legal values: ltr (Default. Left-to-right text direction), rtl (Right-to-left text direction) and auto (let the browser figure out the direction of your alphabet, based on the page content).]]></description>
		  <name>dir</name>
		  <required>false</required>
		  <type>java.lang.String</type>
		</attribute>
		<attribute>
		  <description><![CDATA[Unique identifier of the component in a namingContainer.]]></description>
		  <name>id</name>
		  <required>false</required>
		  <type>java.lang.String</type>
		</attribute>
		<attribute>
		  <description><![CDATA[Boolean value to specify the rendering of the component, when set to false the component will not be rendered.]]></description>
		  <name>rendered</name>
		  <required>false</required>
		  <type>java.lang.Boolean</type>
		</attribute>
		<attribute>
		  <description><![CDATA[Inline style of the input element.]]></description>
		  <name>style</name>
		  <required>false</required>
		  <type>java.lang.String</type>
		</attribute>
		<attribute>
		  <description><![CDATA[Style class of this element.]]></description>
		  <name>style-class</name>
		  <required>false</required>
		  <type>java.lang.String</type>
		</attribute>
		<attribute>
		  <description><![CDATA[Style class of this element.]]></description>
		  <name>styleClass</name>
		  <required>false</required>
		  <type>java.lang.String</type>
		</attribute>
		<attribute>
		  <description><![CDATA[The text of the tooltip.]]></description>
		  <name>tooltip</name>
		  <required>false</required>
		  <type>java.lang.String</type>
		</attribute>
		<attribute>
		  <description><![CDATA[Where is the tooltip div generated? That's primarily a technical value that can be used to fix rendering errors in special cases. Also see data-container in the documentation of Bootstrap. The default value is body.]]></description>
		  <name>tooltip-container</name>
		  <required>false</required>
		  <type>java.lang.String</type>
		</attribute>
		<attribute>
		  <description><![CDATA[Where is the tooltip div generated? That's primarily a technical value that can be used to fix rendering errors in special cases. Also see data-container in the documentation of Bootstrap. The default value is body.]]></description>
		  <name>tooltipContainer</name>
		  <required>false</required>
		  <type>java.lang.String</type>
		</attribute>
		<attribute>
		  <description><![CDATA[The tooltip is shown and hidden with a delay. This value is the delay in milliseconds. Defaults to 0 (no delay).]]></description>
		  <name>tooltip-delay</name>
		  <required>false</required>
		  <type>java.lang.Integer</type>
		</attribute>
		<attribute>
		  <description><![CDATA[The tooltip is shown and hidden with a delay. This value is the delay in milliseconds. Defaults to 0 (no delay).]]></description>
		  <name>tooltipDelay</name>
		  <required>false</required>
		  <type>java.lang.Integer</type>
		</attribute>
		<attribute>
		  <description><![CDATA[The tooltip is hidden with a delay. This value is the delay in milliseconds. Defaults to 0 (no delay).]]></description>
		  <name>tooltip-delay-hide</name>
		  <required>false</required>
		  <type>java.lang.Integer</type>
		</attribute>
		<attribute>
		  <description><![CDATA[The tooltip is hidden with a delay. This value is the delay in milliseconds. Defaults to 0 (no delay).]]></description>
		  <name>tooltipDelayHide</name>
		  <required>false</required>
		  <type>java.lang.Integer</type>
		</attribute>
		<attribute>
		  <description><![CDATA[The tooltip is shown with a delay. This value is the delay in milliseconds. Defaults to 0 (no delay).]]></description>
		  <name>tooltip-delay-show</name>
		  <required>false</required>
		  <type>java.lang.Integer</type>
		</attribute>
		<attribute>
		  <description><![CDATA[The tooltip is shown with a delay. This value is the delay in milliseconds. Defaults to 0 (no delay).]]></description>
		  <name>tooltipDelayShow</name>
		  <required>false</required>
		  <type>java.lang.Integer</type>
		</attribute>
		<attribute>
		  <description><![CDATA[Where is the tooltip to be displayed? Possible values: "top", "bottom", "right", "left", "auto", "auto top", "auto bottom", "auto right" and "auto left". Default to "bottom".]]></description>
		  <name>tooltip-position</name>
		  <required>false</required>
		  <type>java.lang.String</type>
		</attribute>
		<attribute>
		  <description><![CDATA[Where is the tooltip to be displayed? Possible values: "top", "bottom", "right", "left", "auto", "auto top", "auto bottom", "auto right" and "auto left". Default to "bottom".]]></description>
		  <name>tooltipPosition</name>
		  <required>false</required>
		  <type>java.lang.String</type>
		</attribute>
	</tag>
```

stephanrauh:
One last thought: does it make sense to add support for nested input field? For example, input fields that are put into a `<h:panelGroup>` (aka `<div>`) for some reason?

And do we need to implement extra support for non-horizontal forms?

stephanrauh:
This PR is also related to #614 and #493.

ggam:
Thanks for the review! I had the taglib changed locally but forgot to commit it.

I've added the lines to the constructor, although I don't think it's necessary as it will anyway call the parent constructor that already does it.

As for you latest questions:
*  Support for `h:panelGrid`: how would you handle that? setting an attribute on the panelGrid to instruct inner inputs not to generate use the `form-group` class? Instead of checking if the parent component is a FormGroup, we could define a `DONT_GENERATE_FORM_GROUP_ON_INNER_INPUTS` attribute, and check for its presence on the parent component when rendering inputs. That would be provide more flexibility than the current solution. For now, Only FormGroup would have it defined, but I guess we could set it on children `panelGrids` from within the `FormGroupRenderer#encodeChildren`. Anyway, I don't think this is **necessary** at this moment. But if the attribute approach is feasible, we could make use of it in a basic fashion for now and expand it on a future release.
* Support for non horizontal forms: what changes should we need for that? For now, I think just documenting that this component only has sense for horizontal components, and logging a warning when using it on a non-horizontal one would suffice. What are you referring exactly?

stephanrauh:
@ggam I've pushed a couple of bugfixes. I don't expect merge conflicts, but you might want to merge with the last version.

I've also found a bug. The HTML code generated by `<b:selectBooleanCheckbox>` doesn't cope with AJAX. Probably the clientID is rendered in a nested element. It needs to be with the outermost HTML element of the component. You can see the error at http://www3.bootsfaces.net/Showcase/forms/converters.jsf. Just click on "VAT" to see the bug in action.

It's exactly the source code you're already working on, so I don't want to fix it now.

> Support for h:panelGrid: how would you handle that?

Let's simply more the if statement checking for the parent form-group to a new method of the `CoreRenderer`. This way, we can decided how (or if) to implement that feature later.

> Support for non horizontal forms

Throwing a `FacesException` sounds good to me.

ggam:
Adding to CoreRenderer a method that only affects inputs seems dirty to me. What do you think about creating a new InputRenderer that extends from CoreRenderer and adding the method there?

ggam:
It's done now for InputText. Now I have to make it for all the other inputs and test it.

From your list on #682:

b:inputText
b:inputText with appending / prepending facet
b:colorPicker
b:colorPicker with appending / prepending facet
b:selectOneMenu
b:selectOneMenu with appending / prepending facet
b:selectMultiMenu
b:selectMultiMenu with appending / prepending facet
b:datepicker
b:datepicker with appending / prepending facet
b:dateTimePicker
b:dateTimePicker with appending / prepending facet

Also, textarea and I think some more are missing. By the time I do that, could you please review my direction? Thanks!

stephanrauh:
I fine with creating an `InputRenderer` class. I just suggest to call it `CoreInputRenderer` to make its role more obvious and to distinguish it from the `InputRenderer` of PrimeFaces.

stephanrauh:
> although I don't think it's necessary as it will anyway call the parent constructor that already does it.

Only if you call it :). The original implementation didn't include the `super()` call.

stephanrauh:
I think the second ` beginDisabledFieldset(formGroup, rw);` should be an `endDisabledFieldset()`.

stephanrauh:
I'm not sure Javadoc copes with your HTML formatting. In any case, it shows two error messages:
```
DateTimePickerRenderer.java:88: error: @param name not found
	 * @param format
	          ^
InputTextRenderer.java:61: error: @param name not found
     * @param The real attribute name of the request parameter. By default, BootsFaces guesses the attribute name
```
Maybe you can have a look at the second error because it's a class you're already working on. I'd like not to have too many merge conflicts :).

stephanrauh:
I've modified the DateTimePicker, the ColorPicker and the Sliders slightly. Probably still no merge conflict, but pulling the current master might be a good idea.

ggam:
I've changed the class to CoreInputRenderer. You are right about the beginDisabledFieldSet. I copied it from RowRenderer. I've fixed it for both and I pulled from master. No conflicts for now!

HTML in JavaDoc is ugly to read in code, but outputs formatted text. I can remove it if you prefer..

About the constructor, as you now, Java will always call the default parent constructor if no other is specified ;)

I'm now going to work on updating the rest of the component. Hopefully I'll be able to do it today.

ggam:
I think I have it!

I've tested a basic version of the following components:

b:inputText
b:colorPicker
b:selectOneMenu
b:selectMultiMenu
b:datepicker
b:dateTimePicker
b:inputTextArea
b:selectOneCheckbox

At the same time, two bugs where fixed:
- ColorPicker was not rendering the label by default
- DatePicker didn't declared the label attribute. I think the getAttribute method is no present in any on the interfaces implemented by inputs. This should be checked and defined.

Could you please review this and merge if everything is ok?

ggam:
Btw, this issue is tagged as "bug", but is really an enhancement and nothing was broken

stephanrauh:
That was quick. Thank you very much for the good work!

stephanrauh:
What about #614 and #493? Are they solved now, too?

stephanrauh:
> I think the getAttribute method is no present in any on the interfaces implemented by inputs

It's a core method, probably defined in `UIComponent`. But it's old coding style, and we want to get rid of it because we've learned that calling `getAttributes()` is a simple approach, but it doesn't always work. In particular, it often fails with EL expressions. Funny thing is that it doesn't always fail.

Issue #690, opened 2017-04-09
b:form allows horizontal and inline at the same time


ggam:
B:form should throw an exception when both inline an horizontal are set. Now it just add the CSS for both.

I can fix it myself. Posting here so it's not missed

stephanrauh:
The issue has been solved by the PR.

@ggam ¡Muchas gracias!

PR #691, opened 2017-04-09
b:form can't be horizontal and inline at the same time

ggam:
PR for #690.

- Removes FormCore as it was not used.
- Fixes component type and family.


stephanrauh:
`FormCore` is a class automatically generated by the *.jsfdsl generator. The idea being that the core class contains everything that can automatically generated. The `Form` class contains the code that must not be overwritten by the generator. For example, your `encodeBegin()` method belongs to that category.

Issue #675, opened 2017-04-04
Some components render invalid BS markup with col-XX attributes


ggam:
This is a followup from #673.

After testing all the "Bootstrap Widget" components from the showcase, I've found the following failing:
- [x] alert: needs column wrapper
- [x] carousel: needs column wrapper
- [ ] b:dropDown: no visual effect. I guess Bootstrap does not support it. Maybe the attributes could be removed on this component (as they have no effect anyway).
- [x] labels and badges: as reported on #673 
- [x] thumbnail: needs column wrapper
- [ ] tabView: missing col-XX attributes. Should they be added?
- [ ] Tree: haven't tested this. Will need to verify. Noted here just so I don't forget about it.
- [x] Socialshare: missing wrapper.
- [ ] slider2: missing col-XX attributes. Should they be added?
- [ ] canvas: needs column wrapper
- [x] messages: addressed by #715
Interestingly, all the components I've tested from the "Forms" section worked as expected.

stephanrauh:
When I implemented the feature I was mostly interested in forms. But you're right, orthogonality is an important goal.

I don't think we'll manage to solve all this before Easter, so let's postpone it to BootsFaces 1.1.0+.

ggam:
I plan to submit a PR for this soon. Could you please change the milestone to 1.1.0?

stephanrauh:
I'm afraid development of BootsFaces has slowed down considerably. We'll never manage to address this issue. Let's close it.

Issue #673, opened 2017-04-04
b:label and b:badge column sizes generated invalid Bootstrap markup


ggam:
Similar to #670, but now with `<b:label />`.

Example reproducer:
```
<!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/xhtml"
      xmlns:h="http://xmlns.jcp.org/jsf/html"
      xmlns:b="http://bootsfaces.net/ui">
    <h:head>
        <title>Facelet Title</title>
    </h:head>
    <h:body>
        <b:container style="background-color:yellow;">
            <b:row>
                <b:label col-xs="12" text="This doesn't work" style="background-color: blue;" />
            </b:row>
            <b:row style="margin-top: 10px;">
                <b:column col-xs="12">
                    <b:label text="This works" style="display: block;background-color: blue;" />
                </b:column>
            </b:row>
        </b:container>
    </h:body>
</html>
```

![captura](https://cloud.githubusercontent.com/assets/2109040/24654178/e57710da-1938-11e7-9f1e-f8b1f5cc1e94.PNG)

Seems like label needs to be wrapped with a column and `display: block;`. The same happens when using `tinyScreen` and the like attributes.

I'll check the other components to see if they are also affected.

stephanrauh:
Fixed. 

For some reason, the label doesn't seem to need the CSS inline style `display=block`, so I only added it for b:badge.

PR #694, opened 2017-04-10
[WIP] Fixed responsive classes (#675)

ggam:
- [x] alert: needs column wrapper
- [x] carousel: needs column wrapper
- [ ] b:dropDown: no visual effect. I guess Bootstrap does not support it. Maybe the attributes could be removed on this component (as they have no effect anyway).
- [x] labels and badges: as reported on #673 
- [x] thumbnail: needs column wrapper
- [ ] tabView: missing col-XX attributes. Should they be added?
- [ ] Tree: haven't tested this. Will need to verify. Noted here just so I don't forget about it.
- [x] Socialshare: missing wrapper.
- [ ] slider2: missing col-XX attributes. Should they be added?

### DIFF
--- a/src/main/java/net/bootsfaces/component/alert/AlertRenderer.java
+++ b/src/main/java/net/bootsfaces/component/alert/AlertRenderer.java
@@ -64,6 +64,8 @@ public class AlertRenderer extends CoreRenderer {
 		String t = alert.getTitle();
 		boolean closbl = alert.isClosable();
 
+                beginResponsiveWrapper(component, rw);
+                
 		rw.startElement("div", alert);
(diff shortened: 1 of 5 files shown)


stephanrauh:
IMHO tabviews almost always require the full screen width, so the responsive attributes don't make sense.

Slider2 is another issue. That probably makes more sense. I've notices layout glitches with `mode="edit"`, so we need to do something about this, anyways.

stephanrauh:
Do you want us to merge this PR already, or do you want to work on `b:slider2` first?

ggam:
I think this can be merged now. I'll open a new issue for the missing stuff.

 This way testing can start just now.

PR #695, opened 2017-04-10
Use constants for component types

ggam:
Renderers will need to be changed as well but I'll do that in another moment as this was already a lot of work.


stephanrauh:
In general, I've learned that using constants is an antipattern because it hides the real string values. I know, theorists say the contrary. This particular constant is OK, but I wouldn't like to introduce constants for the class names again. It was a lot of work to replace the hard-to-read constants by the actual values.

PR #702, opened 2017-04-14
Improve b:message rendering

### LINKED ISSUE #701: b:message rendering can be improved
- a `null` may appear in the generated `class` attribute
- if there is a `col-` class, wrap the alert in an outer div (see #675 )

I've already prepared a PR

NicolaIsotta:
Fixes #701


NicolaIsotta:
Something in the white spaces is screwing the diff, just add ?w=1 if you want to see a cleaner one

PR #708, opened 2017-04-15
Increase performance of b:tree control 

PokimonZerg:
* Disable node serialization. At now, we don't send serialized node data to the client. Because it decrease performance and have no sense. Previos realization do nothing with this data.
* We send nodeInternalId back from client. So, we can find actual node by this id.
* Pass [] empty array in case there b:tree has no value. (More natural then throwing an exception). We rely to the component realization of empty tree.

(We have faced with preformance issue then we have > 1000 nodes in this tree. So many data sends to client)


stephanrauh:
I've stumbled over the same issue an hour ago. Problem is, AJAX breaks without node serialization. Please double-check if AJAX still works with your PR.

stephanrauh:
Did you remove these lines deliberately, or is it just because you've forked before I added them?

```java 
		Node value = tree.getValue();
		if (null == value) {
			throw new FacesException("Please set the value attribute of b:tree. ClientId of the tree without value: " + clientId);
		}
```

PokimonZerg:
AJAX still works. Tested on our corporate portal page.

Yes, I have removed this exception directly. Replaced with empty JSON array [].

stephanrauh:
Thanks for contributing this PR!

You're lucky: we've just decided to delay the next version of BootsFaces to Monday or to next weekend. So we can merge your last-minute pull request without risk.

stephanrauh:
@PokimonZerg May I mention your name in the "Kudos" section of the BootsFaces 1.1.0 release post?

PR #715, opened 2017-04-20
Improve b:messages rendering

### LINKED ISSUE #714: b:messages rendering can be improved
Just like #701 , the alert should be wrapped in an outer div if a responsive style class is provided.
Also, the onlyMostSevere part isn't working properly right now, because it's based on an unordered Map.

PR coming, no documentation changes needed.

NicolaIsotta:
Fixes #714


NicolaIsotta:
As the other time, something is screwing the diff, just add ?w=1 to the url if you want to see a cleaner one

stephanrauh:
Look OK to me. Thanks!

PR #736, opened 2017-04-26
Fix Ajax call in DateTimePicker ondtchange wrapped in an Outer div

mtvweb:
This fixes Ajax call inside DataTimePicker ondtchange events. Since change of commit TheCoder4eu@821dfca Ajax calls are not executed any more on server side, if DateTimePicker is wrapped in an Outer container.

Example:

	<b:dateTimePicker id="dtp" value="#{ bean.dtpValue }"
				label="DTP Test with AJAX" render-label="true" span="6" show-time="false"
				format="MM.YYYY" tabindex="1" ajax="true" immediate="true" viewMode="months"
				process="@this" update="@form:otherDtp @form:data"
				ondtchange="ajax:bean.updateFilter();" />

The generated Field id is now changed from clientId + "_Input" to "input_" + clientId to fulfill the requirements to be detected in AjaxRenderer

### DIFF
--- a/src/main/java/net/bootsfaces/component/dateTimePicker/DateTimePickerRenderer.java
+++ b/src/main/java/net/bootsfaces/component/dateTimePicker/DateTimePickerRenderer.java
@@ -62,7 +62,7 @@ public class DateTimePickerRenderer extends CoreRenderer {
 			fieldId = clientId + "_Input";
 		}
 		new AJAXRenderer().decode(context, dtp, fieldId);
-//		new AJAXRenderer().decode(context, dtp, clientId);
+		new AJAXRenderer().decode(context, dtp, clientId);
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


mtvweb:
Found another bug. Still not working perfect. Work in progress....

stephanrauh:
I suppose you're up onto something. With all these divs Bootstrap wraps around input fields, it's easy to use the wrong id, thus breaking our AJAX engine.

mtvweb:
At least it is somewhere between the id used for the "Outer" div. The dp.change is bound to this, but AjaxRenderer decode doesn't find the bound "Ajax:action", because it can't find the reference in "form:dtpOuter". Not sure yet, if I need to fix it in DateTimePickerRenderer decode or AjaxRenderer.

PR #737, opened 2017-04-27
Fix Ajax call in ondtchange of DateTimePicker

mtvweb:
This fixes the passed id  to AJAXRenderer decode-call, if the DateTimePicker is rendered in an Outer-Container and dtchange event is raised. Also it fixes dtchange event with a field-id set.

If a dtchange event is raised, which is bound to the outer div it passes the outer id, else it  uses the default behavior (fieldId or clientId + "_Input"). i.E. a blur event is rendered with the field-id set.

Example:

```xml
<b:dateTimePicker id="dtp" value="#{ bean.dtpValue }"
				label="DTP Test with AJAX" render-label="true" span="6" show-time="false"
				format="MM.YYYY" tabindex="1" ajax="true" immediate="true" viewMode="months"
				process="@this" update="@form:otherDtp @form:data"
				ondtchange="ajax:bean.updateFilter();" />
```


stephanrauh:
@mtvweb Is your PR ready to be merged? And what does it do precisely? As far as I can see, it does two things: updating the dateTimePicker CSS and JS to a newer version, and fixing the AJAX bug. Did I get it right?

mtvweb:
I hope that the PR handles now everything well. I've tested it with different configurations and it seems to work well. 

1. it fixes the Ajax bug with datatimepicker components with a label attribute set. These are rendered in an outer container and the event is bound to this container. 
2. The second wasn't a bootsfaces bug, it was a bug in the datetimepicker library. If the datetimepicker was initialized with i.E.

	dayViewHeaderFormat: 'MMMM YYYY', 
	viewMode: 'months', 
	date: moment('01.2017', 'MM.YYYY'), 
	locale: 'de', 
	format: 'MM.YYYY'

it showed the right view on the first open of the datetimerpicker, but if you select a month and open the datetimepicker again it showed the days view (see screenshots). The newer library is handling this well, and show the first view, even on the second selection.

bildschirmfoto 2017-04-29 um 01 36 05
bildschirmfoto 2017-04-29 um 01 36 16

stephanrauh:
Todo left: copy the new version of the css and JS files to the Maven build folder.

PR #755, opened 2017-05-03
Never forbid b:formGroup

### LINKED ISSUE #751: b:formGroup can be used only inside horizontal forms
I think there's no reason to forbid `b:formGroup` inside inline or stacked forms. For example, it could be useful to add some helper text under an input. Example:
help text

NicolaIsotta:
Fix #751


stephanrauh:
@ggam What's your opinion?

ggam:
Seems fine, as commented on #751

PR #764, opened 2017-05-05
Use caption when building the error message

### LINKED ISSUE #763: b:selectBooleanCheckbox: use caption for error message
[Right now](https://github.com/TheCoder4eu/BootsFaces-OSP/blob/e947b4191abfa96a4661df5cd966230410324dfe/src/main/java/net/bootsfaces/component/selectBooleanCheckbox/SelectBooleanCheckboxRenderer.java#L78), when building the error message for a `required b:selectBooleanCheckbox`, only the `label` attribute is used. If it is not provided, `caption` should be checked too, before using the `clientId`.

NicolaIsotta:
Fix #763


stephanrauh:
Looks like a good idea to me. Thanks!

PR #782, opened 2017-05-15
Don't ignore f:selectItem value

### LINKED ISSUE #780: b:selectOneMenu, f:selectItem value is ignored
Consider this code:
```xml
<h:selectOneMenu label="h:selectOneMenu">
	<f:selectItem value="#{index.selectItem}"/>
</h:selectOneMenu>

<b:selectOneMenu label="b:selectOneMenu">
	<f:selectItem value="#{index.selectItem}"/>
</b:selectOneMenu>
```
Bean:
```java
@ManagedBean
@ViewScoped
public class Index {
	private SelectItem selectItem = new SelectItem("DDDDD", "DDDDD");

	public SelectItem getSelectItem() {
		return selectItem;
	}

	public void setSelectItem(SelectItem selectItem) {
		this.selectItem = selectItem;
	}
}
```
The generated HTML is:
```html
<select name="j_idt7" size="1">
	<option value="DDDDD">DDDDD</option>
</select>
<div class="form-group bf-warning has-warning" id="j_idt9">
	<label for="j_idt9Inner" class=" control-label">b:selectOneMenu</label>
	<select id="j_idt9Inner" name="j_idt9Inner" class="form-control">
		<option data-label="&amp;nbsp;">&nbsp;</option>
	</select>
</div>
```
This happens because the `value` attribute of `f:selectItem` is ignored, seems like the code is based on a quite old version of PrimeFaces.

NicolaIsotta:
Fix #780

### DIFF
--- a/src/main/java/net/bootsfaces/component/SelectItemUtils.java
+++ b/src/main/java/net/bootsfaces/component/SelectItemUtils.java
@@ -93,13 +93,16 @@ public class SelectItemUtils {
     }
 
     private static SelectItem toSelectItem(UISelectItem option) {
-		SelectItem item = new SelectItem();
-		item.setDescription(option.getItemDescription());
-		item.setDisabled(option.isItemDisabled());
-		item.setEscape(option.isItemEscaped());
-		item.setLabel(option.getItemLabel());
-		item.setNoSelectionOption(option.isNoSelectionOption());
-		item.setValue(option.getItemValue());
+		SelectItem item = (SelectItem) option.getValue();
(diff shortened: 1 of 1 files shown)


stephanrauh:
Looks like an odd bug. Maybe you're right, it might be a copy and paste bug. I always try to make sure we never copy code from other libraries, but this might have slipped my attention. Worse, it might even be my bug :(.

ErickPezoa:
What version will this fix go in?

PR #788, opened 2017-05-20
Overhauled b:poll

jepsar:
* Component can now be safely updated (previous interval is cleared).
* Added `stop` attribute (to stop poll on update).
* Added `widgetVar` attribute.
* Added `widgetVar.stop()` JavaScript function.
* `update` and `execute` are now processed using `ExpressionResolver`. Note: this might break previous usages (where update and execute were used as is), but this really was a bug. This must be an upgrade note.
* Added `PropertyKeys` and getters and setters.

### DIFF
--- a/build.properties
+++ b/build.properties
@@ -28,7 +28,7 @@ VERSION=1.1.1
 # RELEASE-STATUS - SNAPSHOT , beta1, rc1 ...
 # or don't remove the property but just leave it blank for a Final RELEASE :
 # RELEASE-STATUS=
-RELEASE-STATUS=SNAPSHOT
+RELEASE-STATUS=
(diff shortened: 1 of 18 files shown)


stephanrauh:
Thanks! Funny thing is that the showcase itself was broken by using the `ExpressionResolver`. For some reason unknown, there was a third `b:poll` in the demo. It was never shown, but it didn't matter because of the faulty implementation of the `update` attribute :).

jepsar:
Ah, I now understand what the `once` attribute was supposed to do.

> Component can now be safely updated (previous interval is cleared).

This makes the `once` attribute obsolete.

I saw you already did some work on the documentation. I'll be adding the missing attributes.

PR #791, opened 2017-05-22
Added mask to inputText

jepsar:


stephanrauh:
Looks OK at first glance. Thanks!

Issue #790, opened 2017-05-22
BlockUI always enabled just by importing the library


ggam:
Manually importing the BlockUI library enables it for AJAX requests even with it being disabled on the `web.xml`. BootsFaces AJAX API checks for the presence of the `$.blockUI` and if it's present and not null, it automatically blocks and unblocks the screen: https://github.com/TheCoder4eu/BootsFaces-OSP/blob/5bdbf2288c4cfa06a3c53d1226259e45bfe1708a/gradleResources/staticResources/js/bsf.js#L39

Since BlockUI cannot be disabled on a per request basis (like PrimeFaces does with the `global` attribute), I imported the library and disabled it on `web.xml`. But BootsFaces doesn't do the check on the client side.

I propose two changes here:
- [bug]: importing the library should not automatically enable it
- [feature]: a new way to disable BlockUI for a specific component/event (like PrimeFaces).

The workaround I've used for now is to rename the variables on my imported JS file, downloaded from the author's website.

stephanrauh:
I guess it's easy to implement the check of the web.xml setting. If I'm not mistaken, you've already done similar things. If you want to submit the PR yourself, go ahead!

ggam:
I guess just adding a EL expression to the JS file should work, no?
Something like:

if($.BlockUI != null && #{blockUIEnabled}) {
  // ...
}

The resource mechanism should interpret it. Would that be enough? I'l take
a look and try to provide a PR when I find some time.

chongma:
slightly related: i was doing pagespeed insights https://developers.google.com/speed/pagespeed/insights/ the other day and it said that the BlockUI javascript file was not minified.  does BootsFaces use the minified version?

stephanrauh:
Actually, there's no official minified version of BlockUI. I minified it manually.

stephanrauh:
We've implemented this feature long ago. I've just added the documentation, so we can close the ticket.

PR #792, opened 2017-05-23
Allow to disable BlockUI on a per-request basis

ggam:
This PR allows to disable BlockUI on specific requests when it's globally enabled.

This is related to #790 but doesn't fix it.

Use is as follows:
```
<b:commandButton action="#{index.action}" value="Text" ajax="true">
    <f:param name="blockui.disabled" value="true"/>
</b:commandButton>
```

Should I add the check also to the `BsF.ajax.onevent` function? Should parameters start with revere domain names (i.e.: `net.bootsfaces.blockui_disabled`)?

### DIFF
--- a/gradleResources/staticResources/js/bsf.js
+++ b/gradleResources/staticResources/js/bsf.js
@@ -110,7 +110,10 @@ BsF.ajax.callAjax = function(source, event, update, execute, oncomplete,
 		}
 	}
 	jsf.ajax.request(source, event, opts);
-	if ($.blockUI && $.blockUI != null) {
+	
+	var disableBlockUI = opts['blockui.disabled'] === "true" ? true : false;
+	
+	if ($.blockUI && $.blockUI != null && !disableBlockUI) {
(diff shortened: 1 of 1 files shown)


gradleResources/staticResources/js/bsf.js
```diff
@@ -110,7 +110,10 @@ BsF.ajax.callAjax = function(source, event, update, execute, oncomplete,
 		}
 	}
 	jsf.ajax.request(source, event, opts);
-	if ($.blockUI && $.blockUI != null) {
+	
+	var disableBlockUI = opts['blockui.disabled'] === "true" ? true : false;
+
```
stephanrauh: IMHO you can safely omit the elvis operator. A simple 
```
var disableBlockUI = opts['blockui.disabled'] === "true";
```
does the trick, too :).

ggam: You're right, my statement seems a bit odd.

Issue #803, opened 2017-06-06
b:defaultCommand and b:inputTextarea


chongma:
i set `<b:defaultCommand target="saveDetails" />` inside an `h:form` but it stops me from entering a new line inside a `b:inputTextarea`.  instead it just submits the form

TheCoder4eu:
@chongma the merge seems to have solved this issue, thanks! :-)

PR #804, opened 2017-06-08
only submit form on input

chongma:
#803 small syntax error

### DIFF
--- a/src/main/java/net/bootsfaces/component/defaultCommand/DefaultCommand.java
+++ b/src/main/java/net/bootsfaces/component/defaultCommand/DefaultCommand.java
@@ -67,6 +67,14 @@ public class DefaultCommand extends UIComponentBase {
 						+ "        document.getElementById('" + actionCommandId + "').click();return false; "
 						+ "    } else { " + "        console.log('keycode not 13'); " + "        return true; "
 						+ "    } " + "    }); " + "});", null);
+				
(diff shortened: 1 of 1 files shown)


ggam:
Just to undertand, how does this change the behavior? How will this work from now?

chongma:
it will run the default command defined by `b:defaultCommand` when pressing return on an html `input`.  without the colon if the return key is pressed anywhere in the form, especially `b:inputTextarea` the form would be submitted.  it is the desired behaviour i believe.  just a syntax error in the jquery

chongma:
also i submitted a PR to show a `b:inputTextarea` on the showcase to demonstrate that it works correctly

ggam:
Thanks! I also think that's what is expected. I just wanted to be sure.

Btw, how would control+enter behave? Would that trigger the submit from the text area? That should work in my opinion.

chongma:
could add another line to check for html `textarea` and use check described in this stack overflow https://stackoverflow.com/questions/1684196/ctrlenter-jquery-in-textarea.  i have updated the PR to reflect this

TheCoder4eu:
@chongma , seems ok to me, can I merge it?

chongma:
Yes please.  I tested it but wanted a second opinion before merging

TheCoder4eu:
Ok, thanks, I merged it!

Issue #806, opened 2017-06-12
Disabling BSF-Resources doesn´t work... 


cljk:
I want to use my own bootstrap css and jQuery files since I´ve a ready layout for it (purchased). So I added context-params
```
<context-param>
    <param-name>BootsFaces_USETHEME</param-name>
    <param-value>false</param-value>
</context-param>

<context-param>
    <param-name>net.bootsfaces.get_jquery_from_cdn</param-name>
    <param-value>true</param-value>
</context-param>
<context-param>
    <param-name>net.bootsfaces.get_jqueryui_from_cdn</param-name>
    <param-value>true</param-value>
</context-param>
<context-param>
    <param-name>net.bootsfaces.get_bootstrap_from_cdn</param-name>
    <param-value>true</param-value>
</context-param>
<context-param>
    <param-name>net.bootsfaces.get_fontawesome_from_cdn</param-name>
    <param-value>true</param-value>
</context-param>
```
But the header tags of BSF are still generated. Is it possible to disable them all or did I configure it wrong?

```
<?xml version='1.0' encoding='UTF-8' ?>
<!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/xhtml">
   <head id="j_idt2">
      [... my CSS tags & prime faces...]
	  
	  <link type="text/css" rel="stylesheet" href="/javax.faces.resource/css/default/core.css.xhtml?ln=bsf" />
      <link type="text/css" rel="stylesheet" href="//maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css" />
      <link type="text/css" rel="stylesheet" href="/javax.faces.resource/css/bsf.css.xhtml?ln=bsf" />
      <!--[if lt IE 9]><script src="/javax.faces.resource/js/html5shiv.js.xhtml?ln=bsf"></script><script src="/javax.faces.resource/js/respond.js.xhtml?ln=bsf"></script><![endif]-->
	  
	  <script type="text/javascript" src="/javax.faces.resource/jquery/jquery.js.xhtml?ln=primefaces&amp;v=6.1"></script>
	  <script type="text/javascript" src="/javax.faces.resource/core.js.xhtml?ln=primefaces&amp;v=6.1"></script>
	  <script type="text/javascript" src="/javax.faces.resource/components.js.xhtml?ln=primefaces&amp;v=6.1"></script>
[...]

PR #809, opened 2017-06-19
#806 Disabling BSF-Resources doesn´t work

cljk:
Addressing issue #806 to be able to load own version of Bootstrap - like described in https://showcase.bootsfaces.net/miscellaneous/Configuration.jsf ("Use your own version of jQuery, jQueryUI, FontAwesome, or Bootstrap instead of using the version BootsFaces brings").

Demo that fix is working: https://github.com/cljk/bsf-issue-806-example

Open issue: CSS of components still get added (perhaps to be fixed later). bsf.css is still added - which should be correct (and modified in the documentation).

Changes:
* move invoking removeBootstrapResources() to end of addCSS() (CSS files were removed and added again)
* fixed logic of "load Bootstrap from CDN" in addCSS() and renamed flag (compare to load jQuery logic)
* added position "last" to "bsf.js" to be able to add custom jQuery before (bsf.js depends on it)
* added named constants for paremeters

### DIFF
--- a/src/main/java/net/bootsfaces/C.java
+++ b/src/main/java/net/bootsfaces/C.java
@@ -36,9 +36,26 @@ public final class C {
 
     //Theme
     public static final String BSF_LIBRARY="bsf";
+    
     public static final String P_USETHEME ="BootsFaces_USETHEME";
     public static final String P_THEME ="BootsFaces_THEME";
 
+    public static final String P_BLOCK_UI = "net.bootsfaces.blockUI";
+    
+	public static final String P_GET_JQUERYUI_FROM_CDN = "net.bootsfaces.get_jqueryui_from_cdn";
+	public static final String P_GET_JQUERY_FROM_CDN = "net.bootsfaces.get_jquery_from_cdn";
+	public static final String P_GET_BOOTSTRAP_FROM_CDN = "net.bootsfaces.get_bootstrap_from_cdn";
+	public static final String P_GET_FONTAWESOME_FROM_CDN = "net.bootsfaces.get_fontawesome_from_cdn";
+
+	// public static final String P_GET_BOOTSTRAP_COMPONENTS_FROM_CDN = "net.bootsfaces.get_bootstrap_components_from_cdn";
+
+	
+	
+
+	public static final String THEME_NAME_DEFAULT = "default";
+	public static final String THEME_NAME_OTHER = "other";
+
+    
     //Font Awesome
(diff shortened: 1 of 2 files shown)


stephanrauh:
You mentioned you've renamed the flag `net.bootsfaces.get_bootstrap_from_cdn`.  However, as far as I can see, the name of the flag is the same, but you've inverted its meaning. Did I miss something?

TheCoder4eu:
@cljk , @stephanrauh , the statement 
`loadBootstrap = shouldLibraryBeLoaded(P_GET_BOOTSTRAP_FROM_CDN, true);`

assumes by default it is true we want to load Bootstrap from CDN instead than from the library.
For this reason I liked more the name of the variable loadBootstrap to be loadBootstrapFromCDN.
Hence the statement 
`if (loadBootstrapFromCDN) { removeBootstrapResources(root, context); }`
made perfectly sense, so I agree the semantic of the parameter is inverted in the PR.

stephanrauh:
I don't mind the inversion as such - I'm just concerned about backward compatibility. Does this change break existing programs? If so, we should call the next version of BootsFaces 1.2.0, not 1.1.2. Plus, we should shout the breaking change out load and clear.

cljk:
Sorry for the late reply ... 

* I didn´t rename the parameter `net.bootsfaces.get_bootstrap_from_cdn` - but the flag from `loadBootstrapFromCDN` to `loadBootstrap` - because that is what the logic result of `shouldLibraryBeLoaded` actually is. The old name was misleading and source for the adressed bug. You can compare that to the other parameters like `loadJQuery` or `loadJQueryUI` which are used the same (and are not changed). So the *meaning* of the flag is the opposite as before - and it always was.... but... 
* it was non-functional.... because the logic was removing the Bootstrap-CSS .... but this was done in the wrong order and they were added again. Because of that I moved the block to lines 308 ff. It was also non-functional because the resources were removed if `loadBootstrapFromCDN` was true - which was (see above) actually the wrong interpretation of the flag and the reason why I renamed it. So the resources are removed if `! loadBootstrap`.
* extracting the parameter-names to the constant-class C was just cosmetical

Since this patch just enables the feature `net.bootsfaces.get_bootstrap_from_cdn` described in your own doc https://showcase.bootsfaces.net/miscellaneous/Configuration.jsf I don´t think this should break any application. But thats a decision to you.

cljk:
Thank you! ;-)

PR #814, opened 2017-07-09
change minimum interface from List to Iterable

### LINKED ISSUE #813: Feature Request: Allow all kind of iterables as typeahead source
Currently the typeahead-values attribute only supports _List_.

Since the implementation does not use any functionality of a _List_ except the iteration I suppose to change the minimal interface to _Iterable_ to support all kind of iterables or collections.

asbachb:
change the minimum interface to iterable in order to use all kind of iterables as typeahead source

Resolve #813


stephanrauh:
Good idea. Thanks!

Issue #770, opened 2017-05-07
b:slider has a negative indent


stephanrauh:
http://www3.bootsfaces.net/Showcase/jquery-ui/slider.jsf:
![image](https://cloud.githubusercontent.com/assets/3045767/25783077/5d4f9f14-3356-11e7-9e58-1c16ecdadc5a.png)

BonifacioJose:
Is it necessary to include a "row" class at line 136 from SliderRenderer.java? Removing this class apparently solve this issue without any side effects.

![img1](https://user-images.githubusercontent.com/7031658/29535123-14611012-868f-11e7-9f9d-9c6d4fefa812.png)

stephanrauh:
Oh, now I see. We've nested two rows without putting any columns between them. OK, that's clearly not what we intended.

@BonifacioJose Thanks for analyzing the problem and suggesting a solution!

stephanrauh:
Solved by #854.

PR #854, opened 2017-09-04
remove row class at line 136

chongma:
as suggested by @BonifacioJose i removed the row class and now the slider aligns properly in the showcase.  not sure what effect this will have on other operations?


chongma:
this is for #770 by the way

stephanrauh:
Oh, now I see. We've nested two rows without putting any columns between them. OK, that's clearly not what we intended. I'm just surprised nobody detected the error earlier!

stephanrauh:
@chongma Thanks!

PR #870, opened 2017-09-24
Fix for responsive Panel and small fix for Mojarra EL

mtvweb:
1. This fixes non collapsible panel with responsive definition. i.E.

`<b:panel title="Title" span="6" collapsible="false">`

2.  it fixes also (small hack for) the Mojarra expression, if the value isn't resolvable for value expression. (Was working before commit 7ab7afe)

```
<ui:param name="calculatedAge"
  value="#{ !empty ref ? util.getAgeAsString(ref.birthday) : '' }" />
<b:inputText id="ageDisplay" span="2"
value="#{ calculatedAge }" readonly="true" disabled="true"
renderLabel="true" label="Age" />
```

(The above fakes an input field as display only field for styling purposes)


mtvweb:
To be revised

Issue #848, opened 2017-08-25
<b:dataTable> breaks when using rendered on a column


RuslanVNikolov:
Currently, I've noticed that when using rendered attribute on a <h:column> or  a<b:dataTableColumn> this breaks the page with the following error:
`TypeError: i is undefined[Learn More]  datatables.min.js.jsf:36:271`
This causes the table's design to break. 
Is there a way to do hide a column without this happening? Maybe something I'm not using properly?

stephanrauh:
Like so?

```xml
<b:dataTable value="#{carPool.carPool}" 
              var="car"
              widgetVar="myDataTableWidgetVar">
     <b:dataTableColumn value="#{car.brand}" />
     <b:dataTableColumn value="#{car.type}" rendered="false"/>
     <b:dataTableColumn value="#{car.color}" rendered="false"/>
     <b:dataTableColumn value="#{car.year}"  order="asc" rendered="false" />
     <b:dataTableColumn value="#{car.price}"  rendered="false"/>
     <b:dataTableColumn value="#{car.enginePower}" label="Engine Power (hp)" rendered="false" />
</b:dataTable>```

It works fine on my machine. Only when I set `rendered="false"` on every column, there's an error message. Even so, it a different error message.

So what does your code look like exactly?

stephanrauh:
@RuslanVNikolov I'll close the ticket now because I can't reproduce it. If you send us a reproducer, we'll happily have a look at it. It's just that I don't have a clue how to proceed because my tests didn't show any problems.

vsvetoslavov:
Hi
I confirm we have this problem (plus I work together with Ruslan). I did a little bit of investigation and noticed that there is a column list (something like `columns: [null, null, null, ..., {orderable: false}]` ) inside the generated javascript. Further investigation led me to the following lines inside DataTableRenderer class:
`String orderable = column.getAttributes().get("orderable").toString();
if ("false".equalsIgnoreCase(orderable)) {
    if (dataTable.getColumnInfo() == null) {
        List<String> infos = new ArrayList<String>(dataTable.getChildren().size());
        for (int k = 0; k < dataTable.getChildren().size(); k++) {
            infos.add(null);
        }
        dataTable.setColumnInfo(infos);
    }
    List<String> infos = dataTable.getColumnInfo();
    String s = infos.get(index);
    if (s == null) {
        infos.set(index, "'orderable': false");
    } else
[...]

PR #875, opened 2017-09-28
Fix for dataTable regading columns with rendered="false"

vsvetoslavov:
Made the generation of column definitions respect the children being rendered or not - if there are column definitions for columns that are not rendered this causes a JavaScript error.
Issue: #848


stephanrauh:
@vsvetoslavov First of all, please apologize we didn't review your pull request earlier. These things shouldn't happen - but obviously, they do happen. Thank you very much for your pull request!

stephanrauh:
As for the PR itself: looks fine by me. It's interesting to learn I've repeated the same code no less than four times. That, in turn, helped to obscure the bug. Let's merge the PR!

vsvetoslavov:
@stephanrauh thanks :)

PR #909, opened 2018-02-05
allow Integer itemValue for selectMultiMenu besides String only

jottyfan:
Currently, only string values are allowed for selectMultiMenu items. A typical use case for such a component is to have a description in the label and an Integer in the value. Therefore, I added a cast to Integer for the itemValue and convert it to a String, so that the following construction should output a list of itemValues instead of a list of index values:

```java
public class MyBean() {
  private final Integer key;
  private String value;
  ...
}

public List<MyBean> myList;
...
public String myValue;
...
```
```xml
<b:selectMultiMenu value="#{myValue}">
  <f:selectItems value="#{myList}" var="l" itemValue="#{l.key}" itemLabel="#{l.value}" />
</b:selectMultiMenu>
```
One could argue that not only Integer, but also Float or Boolean can become important - well, the solution is easy. But imho, Integer ist much more important as I use this on database level as a primary key - description pair, mostly the primary key is from a generated index (and this is a number).

For pairs such as

| pk | description  |
|----|--------------|
| 10 | first choice |
| 23 | none of them |

I don't want `myValue` to have `0,1` as values if I select everything, but `10,23` instead.

### DIFF
--- a/src/main/java/net/bootsfaces/component/selectMultiMenu/SelectMultiMenuRenderer.java
+++ b/src/main/java/net/bootsfaces/component/selectMultiMenu/SelectMultiMenuRenderer.java
@@ -637,6 +637,9 @@ public class SelectMultiMenuRenderer extends CoreInputRenderer {
 			String value;
 			if (itemValue instanceof String) {
 				value = (String) itemValue;
+			} else if (itemValue instanceof Integer) {
+				Integer intValue = ((Integer) itemValue);
+				value = intValue == null ? null : intValue.toString();
 			} else
 				value = String.valueOf(index);
 			rw.writeAttribute("value", value, "value");


stephanrauh:
Thanks!

I guess b:selectOneMenu could benefit from the same improvement.

jottyfan:
You're welcome. In fact, the `<b:selectOneMenu />` still accepts Integer keys, I'm using this very much for years. This is the first time that I'm using `<b:selectMultiMenu />`, so this PR is just for that.

stephanrauh:
I've upload a new developer snapshot of BootsFaces 1.2.1 to Maven Central. As usual, have a look at #369 on how to get it. 

Would you mind to test it? Thanks in advance!
Stephan

jottyfan:
Seems not to work, this is the error message: `Invalid dataCouldn't set the value of the b:selectMultiMenu because an option that wasn't defined by the JSF source code was passed.`

I think the problem is a conversion thing on loading the values from the list when a filled multi menu is read. Currently, I'm out of time to fix this, maybe you can have a look at that and find out where the solution could be found? Somewhere, a string parsing routine of `1,3,99`must be done to convert it back to the select fields inside of the dropdown.

PR #919, opened 2018-03-05
New Component b:breadcrumbs and Fix to selectOneMenu with converter

mtvweb:
This pull request contains the following changes:

1. New Component **_b:breadcrumbs_**

Example:

```xhtml
<b:breadcrumbs>
  <b:navLink ....>
  <b:navCommandLink ...>
</b:breadcrumbs>
```

2. Fix to b:selectOneMenu with Converter. Tested with
(Example taken from : https://memorynotfound.com/using-custom-converter-for-hselectonemenu/)

```java
public class Beer {

    private Integer id;
    private String brand;

    public Beer(Integer id, String brand) {
        this.id = id;
        this.brand = brand;
    }

    public Integer getId() {
        return id;
    }

    public void setId(Integer id) {
        this.id = id;
    }

    public String getBrand() {
        return brand;
    }

    public void setBrand(String brand) {
        this.brand = brand;
    }
}
```

```java
@Named
@ViewScoped`
public class BeersBean implements Serializable {

	/**
	 * 
	 */
	private static final long serialVersionUID = -3192521384162408966L;
	private Beer selectedBeer;
	private List<Beer> beers;

	public BeersBean() {
	}
	
	@PostConstruct
	public void _init() {
		beers = new ArrayList<Beer>();
		beers.add(new Beer(10, "La Chouffe"));
		beers.add(new Beer(20, "Stella Artois"));
		beers.add(new Beer(30, "Westmalle Trippel"));
	}

	public Beer getSelectedBeer() {
		return selectedBeer;
	}

	public void setSelectedBeer(Beer selectedBeer) {
		System.out.println("set:" + selectedBeer);
		this.selectedBeer = selectedBeer;
	}

	public List<Beer> getBeers() {
		return beers;
	}

	public void setBeers(List<Beer> beers) {
		this.beers = beers;
	}

	public Beer getBeer(Integer id) {
		if (id == null) {
			throw new IllegalArgumentException("no id provided");
		}
		for (Beer beer : beers) {
			if (id.equals(beer.getId())) {
				return beer;
			}
		}
		return null;
	}
}
```

```java
@FacesConverter(value = "beerConverter")
public class BeerConverter implements Converter {

    @Override
    public Object getAsObject(FacesContext ctx, UIComponent uiComponent, String beerId) {
        ValueExpression vex =
                ctx.getApplication().getExpressionFactory()
                        .createValueExpression(ctx.getELContext(),
                                "#{beersBean}", BeersBean.class);

        BeersBean beers = (BeersBean)vex.getValue(ctx.getELContext());
        return beers.getBeer(Integer.valueOf(beerId));
    }

    @Override
    public String getAsString(FacesContext facesContext, UIComponent uiComponent, Object beer) {
        return ((Beer)beer).getId().toString();
    }

}
```

3. Enabled Option "hideNoSelectOption" for b:selectOneMenu

4. Modified Loop in navLink, navCommandLink to find parent boots faces component


chongma:
Thanks looks good! This fixes converters on selectOneMenu. I will try to test it shortly

stephanrauh:
@chongma If you'd like to merge the PR, go ahead! I've briefly scanned the changes, and I've exchanged a couple of e-mails with @mtvweb, so I'm positive we can merge the PR without further ado. But still, it has to be tested. What's your time-table?

chongma:
I pulled @mtvweb fork master and tried it with my converter test project but there were errors. It said `f:converter` was not allowed as child of `b:selectOneMenu`. I removed that part and the page rendered badly. Looking at html it had bootsfaces namespace in the main html tags. It may be that I am doing something wrong. I was going to take another look today

chongma:
first error was `javax.faces.view.facelets.TagException: /converter.xhtml at line 68 and column 54 <f:converter> Parent not composite component or an instance of ValueHolder: javax.faces.component.html.HtmlForm@7205c2d4
	org.apache.myfaces.view.facelets.tag.jsf.ConverterTagHandlerDelegate.apply(ConverterTagHandlerDelegate.java:100)` and as i said the bootsfaces namspace is in the html tag
2018-03-07 12 21 33 pm
and then it looks like this
2018-03-07 12 22 31 pm
whereas with 1.2.0 stable it looks like this
2018-03-07 12 24 49 pm
should the @mtvweb master fork work similarly to the bootsfaces master branch?

mtvweb:
Added my test classes, xhtml and build bootsfaces.jar as a zip file.

[ConvertTest.zip](https://github.com/TheCoder4eu/BootsFaces-OSP/files/1789397/ConvertTest.zip)

chongma:
yes it works with the packaged jar.  the converter also seems to be working which is awesome

stephanrauh:
@chongma Can we merge the PR? I'm a bit confused because your comments seem to indicate that the merged version doesn't work, while the binary version Jens sent to you does work. Did I get this right?

chongma:
@stephanrauh I struggled with the forked project but the zipped jar worked fine. Is there some way i can clone and test the PR before merge? Otherwise we can just merge and see what happens?

mtvweb:
I‘ve build my file from mtvweb Master with

gradlew build
mvn install

Meanwhile i‘ve äffend the mavenfiles generated by gradle too, but These only affects mainly css files.

mtvweb:
BTW: It seems that the select2 files are disappearing somewhere during the gradle / mvn build

mtvweb:
Sorry, it seems that GitHub is not working properly with iPhone Safari. Some strange things are happen there.

stephanrauh:
@chongma I had the same difficulties with GitHub. I followed the instructions to merge the PR locally. Git insisted an merging the changes and committing a merge commit to the branch of Jens. In the end, I did just this and checked the files individually.

stephanrauh:
I've merged the PR and added the beer demo to the showcase. I just wonder if we should add some Spanish, Italian, and German beer brands :).

stephanrauh:
I've also fixed the gradle build. Now it includes the select2 files. Thanks for pointing this out, @mtvweb 
.

stephanrauh:
I've added a documentation page for the breadcrumb component. @mtvweb Do you have any suggestion what to add to the page?

BTW: I'll add the reference section later. It's generated automatically, but currently I've got the wrong version of the plugin installed, so it can only generate code for BootsFaces 2.

mtvweb:
In fact it's only another kind of a link list, like b:listLinks, b:navLinks, b:tabLinks or b:pillLinks. So I don't see nothing special to document. A standard navLink or navCommandLink should render well in the new component. Separater will result in an empty entry "/ /"

PR #959, opened 2018-07-03
openstreetmap

chongma:


stephanrauh:
@TheCoder4eu Do we want to include this PR with BootsFaces 1.3.0? If so, I could run a test tomorrow evening. What's your opinion?

BTW, the PR https://github.com/TheCoder4eu/BootsFacesWeb/pull/56 also belongs to this change.

@chongma Did you modify the BootsFaces.jsfdsl, too? If not, we should copy the changes into that file, too, because it generates all the boilerplate code of the component and the documentation.

chongma:
@stephanrauh no, i did notice the BootsFaces.jsfdsl but wasn't sure it was being used.  i just copied and edited the documentation from another component.  i will take a look

stephanrauh:
@chongma Awesome component! Just one attribute puzzles me. What does `urlTemplate` do?

chongma:
Openstreetmap loads "tiles" which are the squares of map it shows depending on the zoom level. As it is an open platform anyone can host these tiles. So URL template allows you to set the location of the tile provider. We should probably put a note in the documentation that if you abuse the tile provider by using too much bandwidth you will get banned. And if you are a heavy user you need to get permission from the tile provider

chongma:
@stephanrauh maybe we could get listed on this page https://wiki.openstreetmap.org/wiki/Deploying_your_own_Slippy_Map under "Embedding OSM in a CMS/framework"

Issue #963, opened 2018-07-09
dateTimePicker ondtchange broken


jaberu:
ondtchange event does not call backing bean method when not rendered in a div container.
If I render the dateTimePicker with a label everything works well:

```
                <b:dateTimePicker id="thePicker"
                                  value="#{selectDateBean.gueltigkeit}"
                                  format="#{selectDateBean.pattern}"
                                  showTime="false"
                                  process="@this"
                                  update="tableForm"
                                  ondtchange="ajax:selectDateBean.refreshView()"
                                  label="test"
                                  renderLabel="true"
                                  ajax="true">
```

but if I use the dateTimePicker without a label, the backing bean method is not called:

```
                <b:dateTimePicker id="thePicker"
                                  value="#{selectDateBean.gueltigkeit}"
                                  format="#{selectDateBean.pattern}"
                                  showTime="false"
                                  process="@this"
                                  update="tableForm"
                                  ondtchange="ajax:selectDateBean.refreshView()"
                                  ajax="true">
```

This related to the generated component Id generated for the different scenarios.
The AJAXRenderer does not realize the second scenario properly.

jaberu:
Following modification works for me:
PLZ check
https://github.com/TheCoder4eu/BootsFaces-OSP/pull/964

stephanrauh:
As far as I can see, you didn't add a PR for the showcase. Would you mind to submit such a demo, too? That'd help us a lot (and provide an implicit test for each version of BootsFaces to come).

BTW, we've already merged the PR. Thanks!

Best regards,
Stephan

jaberu:
I cannot find the showcase.
Can you post a link to the git hub location?

chongma:
https://github.com/TheCoder4eu/BootsFacesWeb

jaberu:
I checked the showcase and there is already such an example. 

DateTimePicker -> DateTimePicker AJAX

However in the example the label is also rendered. Under those circumstances the error does not appear as I already said. By removing the label attribute the error is reproducable. Does it make sense to add an extra Live Preview and code snippet just without the label attribute?

By the way the compatibility settings button (Labels, required fields and error messages) doesnt seem to work on IE11

PR #964, opened 2018-07-09
fix AJAXRenderer for dateTimePicker

jaberu:
FIX for dateTimePicker ondtchange broken #963


stephanrauh:
I'd like to add this to BootsFaces 1.3.0, but I can't make any promise, because we've already finished testing and preparing the new version. In any case: thanks! I didn't test it yet, but judging from the source code, it most likely works.

stephanrauh:
That doesn't depends on me. My best guess is "this weekend". But currently, both the private and vocational life of every team member is sort of demanding, so I can't promise anything. Except "real soon now".

Issue #989, opened 2018-08-09
navLink and navCommandLink not compatible with keyboard navigation


Lb-lea:
Hello,
it seems that those two items are being skipped during keyboard navigation. 
You can see it on the first example [here]( https://showcase.bootsfaces.net/layout/navbars.jsf ) : when using tab, the selection goes from the brand link to the dropdown one skipping the two in between links. It makes the whole thing un-accessible  for keyboards only users.
It would be great  if there was a fully navigable navbar component.

stephanrauh:
This ticket probably needs some investigation, but you're right, it's one of the things we definitely should support.

Lb-lea:
I'm not sure but from what I gather it might been due to the ```tabindex="-1"``` attribute in the generated link

stephanrauh:
Now keyboard navigation works a lot better. Thanks for reporting the bug and for suggesting the solution!

PR #991, opened 2018-08-10
navlink tabindex -1

chongma:
#989 this seems to fix the problem in the showcase.  not sure if it has any side effects though


stephanrauh:
Oops - sorry, I only saw your pull request after implementing my own bug fix. BTW, my solution is similar to you solution. The difference is that my solution covers b:navLink, b:commandLink, b:dropMenu, b:dropButton, b:navBar, b:link, and b:commandLink. Plus, I saw that we render the tabindex twice (and inconsistently) if it's explicitly defined by the programmer.

PR #1005, opened 2018-09-08
Added float value support to Slider2 fixing #976

### LINKED ISSUE #976: b:slider2 with decimal number (step) values
I would like to be able to slide in steps of 0.25. Is this a feature you are willing to support? Or should I create a composite component for this?

### LINKED ISSUE #988: Expose slider2 JS object to window
Now there is no way to customize things.

https://github.com/TheCoder4eu/BootsFaces-OSP/blob/fa2543991d414eb46d6cb880fc505467f917ba76/src/main/java/net/bootsfaces/component/slider2/Slider2Renderer.java#L376

Could be changed to:

````
"window." + VarName + "_sv =....
````

And maybe even an attribute could be added to set your var name (if your slider is in a form without an id attribute, you get a generated id based variable name like `j_idt8_slider_sv`).

@stephanrauh You could assign this to me if you like.

### PR DESCRIPTION
**Pull request by jepsar**
I've changed the value of the component to Object and detect the type (integer or float). If the type is incorrect an exception will be thrown. This will fix #976 

Please have a look at this first. If this is all OK with you I will work on fixing the precision attribute (I'll create a new issue for that, **update**: #1006) and I will fix #988 

A separate PR will follow for the showcase. **Update**: https://github.com/TheCoder4eu/BootsFacesWeb/pull/58

### CODE REVIEW

File: src/main/java/net/bootsfaces/component/slider2/Slider2Renderer.java [hunk 1]
```diff
@@ -93,35 +80,67 @@ public void encodeBegin(FacesContext context, UIComponent component) throws IOEx
...
 			v = slider.getValue();
 		}
 		if (v == null) {
-			v = slider.getValue();
-			slider.setValue(v);
-		}
-		if (v == null) {
-			v = max / 2;
-			slider.setValue(v);
+			v = min + (max - min) / 2;
```
stephanrauh: That's even a small bug fix, isn't it? :)

jepsar: Yes it was.. see https://github.com/TheCoder4eu/BootsFaces-OSP/issues/976#issuecomment-407347738

PR #1010, opened 2018-09-13
b:modal is not readable by screen reader

Lb-lea:
Hi,
I'm using BootsFaces 1.3.0 with screen reader orca 3.22.2 on firefox 57.0.1
On opening a modal the screen readers reads the title but not the content.
If I try to use keydown to read below, it will key down on the main page and not the modal (thus it reads things that are hidden behind the modal so the user feels very lost).
Using tabs allows to navigate between the title, the cross close button (which would gain to have an aria-label="close"), and the close and ok button. There is no way to read the text content.

It is very weird because the bootstrap one works fine and it is implemented the same way. The only difference I see is that we are missing role="document" on the second level container. Could this be the reason??...


stephanrauh:
@Lb-lea Thanks for you pull request (and sorry for our late response).

Your pull request raises an interesting question. How can we check how the BootsFaces widgets work for people using a screen reader? Is there an open source or free option?

PR #1019, opened 2018-10-20
Added widgetVar to slider2

### LINKED ISSUE #988: Expose slider2 JS object to window

jepsar:
This will fix #988

A PR for the showcase will follow. **Update**: https://github.com/TheCoder4eu/BootsFacesWeb/pull/59

### DIFF
--- a/src/main/java/net/bootsfaces/component/poll/Poll.java
+++ b/src/main/java/net/bootsfaces/component/poll/Poll.java
@@ -17,7 +17,6 @@
  */
 package net.bootsfaces.component.poll;
 
-import javax.faces.context.FacesContext;
 import javax.faces.event.AbortProcessingException;
 import javax.faces.event.ComponentSystemEvent;
 import javax.faces.event.ListenerFor;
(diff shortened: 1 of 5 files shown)


stephanrauh:
Looks good, thanks! Just one suggestion: you've renamed the widget variable. If someone's using this variable, that's a breaking change. So we should either document it as a breaking change, or we should leave the variable name unchanged.

jepsar:
@stephanrauh I only changed the Java name (which did not meet Java standards) that was used in a private method. That won't cause any problems will it?

jepsar:
.. ah, if you mean the "_sv" part, that was moved. So the name is still the same:

https://github.com/jepsar/BootsFaces-OSP/blob/a6b5f2e1c59c7f4cdad2a29d3f52e1aac8602aad/src/main/java/net/bootsfaces/component/slider2/Slider2Renderer.java#L389

PR #1032, opened 2018-11-16
add <caption> optional tag to datatable

### LINKED ISSUE #1028: Datatable: adding the option to add a  <caption>
Hello,
Table captions and summaries provide information to give the user a quick information of the table content. While they are not required to meet WCAG 2.0, they can still be very useful for screen reader users.
Usage example
```
<table>
  <caption>Explaining what this table is about</caption>
  <tr>...</tr>
</table>
```
or
```
<table   summary="Explaining what this table is about">
```

It might be interesting to add those or one of those optional field in the datatable.

Lb-lea:
add <caption> optional tag to datatable to improve screen reader experience

Fix  #1028

### DIFF
--- a/src/main/java/net/bootsfaces/component/dataTable/DataTableCore.java
+++ b/src/main/java/net/bootsfaces/component/dataTable/DataTableCore.java
@@ -24,7 +24,7 @@ import javax.faces.component.UIData;
 public abstract class DataTableCore extends UIData implements net.bootsfaces.render.IHasTooltip {
 
 	protected enum PropertyKeys {
-		ajax, autoUpdate, border, colLg, colMd, colSm, colXs, columnVisibility, contentDisabled, copy, csv, customLangUrl, customOptions, delay, deselectOnBackdropClick, disabled, display, excel, fixedHeader, hidden, immediate, info, lang, largeScreen, markSearchResults, mediumScreen, multiColumnSearch, multiColumnSearchPosition, offset, offsetLg, offsetMd, offsetSm, offsetXs, onclick, oncomplete, ondblclick, ondeselect, onerror, onmousedown, onmousemove, onmouseout, onmouseover, onmouseup, onorder, onpage, onsearch, onselect, onsuccess, pageLength, pageLengthMenu, paginated, pdf, print, process, responsive, rowGroup, rowHighlight, rowStyleClass, saveState, scrollCollapse, scrollHorizontally, scrollSize, scrollX, searching, select, selectedColumn, selectedItems, selectedRow, selectionInfo, selectionMode, smallScreen, span, striped, style, styleClass, tinyScreen, tooltip, tooltipContainer, tooltipDelay, tooltipDelayHide, tooltipDelayShow, tooltipPosition, update, visible, widgetVar;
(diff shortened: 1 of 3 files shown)


stephanrauh:
Awesome, thanks! First impressions, without having run the code yet:

1) Awesome, thanks! :)
2) Please add the caption attribute to the BootsFaces.dsl. I can do that for you, but if you do, you make our lifes easier. Also see https://github.com/stephanrauh/JSFLibraryGenerator.
3) Your approach is great for the `summary` attribute of the datatable. If I've got it right, the `<caption>` tag offers a bit more flexibility than the ordinary `summary` attribute. It allows you to use arbitrary HTML tags and (in the case of JSF) even JSF widgets. Implementing this full-featured approach is a bit tedious, so I'm ready to help you. But before I do so, I'd like to ask you if you agree. Should we implement a `<b:caption>` tag that's only valid inside the `<b:datatable>` tag?

Lb-lea:
Your approach is more flexible indeed. It will be closer to the way html tables are declared as well so it is more intuitive to use. 
I'd be really interested to learn how to make the new features if you don't mind helping me. 
I'll take care of the doc accordingly, this should be within my reach ^ ^.
I'm not sure if we need both the summary attribute and the caption. The w3c website advises to have at least one of those.  I could easily switch what I did for the caption into a summary attribute though.

src/main/java/net/bootsfaces/component/dataTable/DataTableCore.java
```diff
@@ -86,6 +86,22 @@ public boolean isBorder() {
...
+	        getStateHelper().put(PropertyKeys.caption, _caption);
+	}
+	
+	
+	/**
+	 * If set, this will surround the table by a border. Defaults to true. <P>
+	 * @return Returns the value of the attribute, or "-1", if it hasn't been set by the JSF file.
+	 */
+	public String getCaption() {
+	        return (String) getStateHelper().eval(PropertyKeys.caption, "-1");
```
jepsar: Why not use `null` here as the default?

Lb-lea: I tried to copy the behaviour of other attributes but it seems I took a size attribute as a reference. You are right I should have used the attributes null  using "-1" doesn't make much sense :disappointed:.

stephanrauh:
@Lb-lea I've decided to speed up things a bit in order to be able to publish BootsFaces 1.4.0 soon. Nonetheless, thanks for offering a PR covering the documentation!

stephanrauh:
Oh, and we certainly appreciate your suggestions, including your pull requests. Actually, we love them. Currently, we're reacting a bit slow because both our jobs and our private life are fairly demanding. Please don't confuse this with negative feedback. If anything, it's us being lazy. Please carry on your good work! Don't hesitate to drop me an email if you need help (webmaster1 at beyondjava.net).

stephanrauh:
@Lb-lea Please review https://staging.bootsfaces.net/Showcase//forms/DataTable.jsf#accessibility. Does the `caption` element work, and did I find the right words?

Lb-lea:
@Lb-lea stephanrauh  works perfectly with orca screen reader ^^. My English is not perfect but I think " is used **to by** screen readers " is weird, shouldn't the " to" be removed?
Anyway don't apologize I'm not very reactive either since I only works on the framework in between classes...

PR #1037, opened 2018-11-25
Add root and end facet to b:breadcrumbs

### LINKED ISSUE #1036: b:breadcrumbs should have an option to render an end point / current page
The end point / current page should not be a link, but can have an icon, so a facet is probably the best solution here. While we're throwing in a facet, we could as well add a facet for a custom root element.

Feel free to assign this to me. I'm happy to provide a PR since it will solve an issue I have with a project of ours.

jepsar:
Fix #1036

See https://github.com/TheCoder4eu/BootsFacesWeb/pull/63


stephanrauh:
IMHO this looks good enough to merge it. I always wince when I see a facet, but this seems one of the cases facets are really useful. Thanks!

jepsar:
Yeah, I thought to create an attribute at first, but that won't allow you to add an icon (for example) without creating more attributes.

PR #1048, opened 2018-12-19
fixed modal accessibility issue

### LINKED ISSUE #1009: b:modal is not readable by screen reader
Hi,
I'm using BootsFaces 1.3.0 with screen reader orca 3.22.2 on firefox 57.0.1
On opening a modal the screen readers reads the title but not the content. 
If I try to use keydown to read below, it will key down on the main page and not the modal (thus it reads things that are hidden behind the modal so the user feels very lost). 
Using tabs allows to navigate between the title, the cross close button (which would gain to have an **aria-label="close"**), and the close and ok button. There is no way to read the text content.

It is very weird because the bootstrap one works fine and it is implemented the same way. The only difference I see is that we are missing **role="document"**  on the second level container. Could this be the reason??...

Lb-lea:
removed the aria-hidden=true that made the modal invisible for the screen reader.
Fix #1009 

I found what was really causing the issue with the screen reader ~:star:
the aria-hidden=true is still there when the modal is open that is why it gets ignored! 

The problem is that aria-hidden was introduced in the bootstrap 4 modal but we are using the bootsrap 3 modal.js.
If BF stays with bootsraps 3 we can't use the aria-hiddenfor this component.

### DIFF
--- a/src/main/java/net/bootsfaces/component/modal/ModalRenderer.java
+++ b/src/main/java/net/bootsfaces/component/modal/ModalRenderer.java
@@ -55,13 +55,13 @@ public class ModalRenderer extends CoreRenderer {
 		}
 		/*
 		 * <div id="myModal" class="modal fade" tabindex="-1" role="dialog"
-		 * aria-labelledby="myModalLabel" aria-hidden="true"> <div
+		 * aria-labelledby="myModalLabel" > <div
 		 * class="modal-dialog"> <div class="modal-content"> <div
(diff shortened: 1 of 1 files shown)


stephanrauh:
That's a surprise! We copied the `aria-hidden` attribute directly from the Bootstrap documentation (https://bootstrapdocs.com/v3.3.6/docs/javascript/#modals). But I trust you and merge the PR nonetheless.

Issue #999, opened 2018-08-25
defaultCommand does not work on commandButton with col/span attribute


TheTimeWalker:
After trying to fight forever on why `<b:defaultCommand/>` didn't work on `<b:commandButton/>`, I've finally figured out the culprit

Steps to reproduce:
1. Set up a form with a generic inputText
2. Add a commandButton with any of these attributes: colXs, colSm, colMd, colLg, span
3. Bind defaultCommand with a target to commandButton

What happens is that the button is being wrapped around a `<div>` which inherits the ID of the button. The button itself contains no ID. This means that defaultCommand instead clicks on the div as opposed to the button.

Tested on:
BootsFaces 1.3.0
PrimeFaces 6.2.0
Payara 5.181 with JSF 2.3

stephanrauh:
Oops. Thanks for pointing this out!

geopossachs:
i retest it, the bug was fixed with version 1.4.2

PR #1075, opened 2019-05-06
use document.getElementsByName() to trigger default command #999

geopossachs:
use document.getElementsByName()[0] to trigger the default command by a commandButton with col/span attribute #999


stephanrauh:
I've added an example to the showcase demonstrating how your improvement works.

stephanrauh:
Thanks!

I'm not entirely sure how your improvement works with PrimeFaces buttons, standard JSF button, and `<b:commandLink>` - but then, we never documented these features are supported :)

PR #1083, opened 2019-06-12
Fix value submission on InputText with tags

### LINKED ISSUE #1079: b:inputText not working in Chrome
I'm trying to write multiple tags to a managed bean, but the value is always an empty string.
My InputText looks like:

`                    
<b:inputText id="dataTypeProcedures"
                                 value="#{dataTypeBean.tags}"
                                 placeholder="Verfahren"
                                 tags="true"
                                 typeahead="true"
                                 typeahead-values="#{dataTypeBean.avaliableProcedures}"
                                 tiny-screen="full-width" />`

I think Problem is, that the input field is rendererd twice with the same name in html, why Chrome sents two values back to server, one empty, one with correct. However the empty value is set to the backing bean.
image
I'm using Chrome Version 70.0.3538.77 , Bootsfaces 1.4.1

pkelbert:
Tagsinput clones InputText 'input' HTML element and keeps both original
and newly created elements with 'name' attribute.

This commit removes duplicated 'name' attribute on newly created HTML
'input' element via inline JavaScript for each InputText component with
tags enabled.

Fix #1079

### DIFF
--- a/src/main/java/net/bootsfaces/component/inputText/InputTextRenderer.java
+++ b/src/main/java/net/bootsfaces/component/inputText/InputTextRenderer.java
@@ -232,9 +232,6 @@ public class InputTextRenderer extends CoreInputRenderer {
 		if ((autocomplete != null) && (autocomplete.equals("off"))) {
 			rw.writeAttribute("autocomplete", "off", null);
 		}
-		if (inputText.isTags() && (!inputText.isTypeahead())) {
-			rw.writeAttribute("data-role", "tagsinput", null);
-		}
 
 		String v = getValue2Render(context, component);
 		if (inputText instanceof InputSecret) {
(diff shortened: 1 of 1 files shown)


pkelbert:
Application used for testing: https://github.com/pkelbert/bootsfaces-inputtags-demo

stephanrauh:
@pkelbert Sounds reasonable to me. Thanks! Nonetheless, I've got a question before merging the PR. Why did you remove the attribute `data-role="tagsinput"`?

pkelbert:
@stephanrauh The attribute `data-role="tagsinput"` was removed in order to initialize the component manually instead of automatically. When the component is initialized automatically I couldn't find any safe criteria for selecting the original HTML element once Tags Input removes it's `id` attribute.

stephanrauh:
@pkelbert Obrigado!

pkelbert:
@stephanrauh Thanks!

Issue #1093, opened 2019-08-31
Upgrade Bootstrap to v3.4.1 and Bootswatch themes to v3.4.1+1 and test.


TheCoder4eu:
Upgrade Bootstrap to v3.4.1 and Bootswatch themes to v3.4.1+1 and test.

mtvweb:
With 3.4.1 the filter field of the data table looks strange. (But seems to be a DataTables problem)
Even the b:selectOneMenu text is not positioned well. At least with field-size="sm"

geopossachs:
hello @mtvweb that's wonderful that you want to support the project.

i would like to review your changes. It is possible to shorten the pull request to the central changes here? (rebase; chery pick; ...)

geopossachs:
> Even the b:selectOneMenu text is not positioned well. At least with field-size="sm"
#1097

stephanrauh:
@mtvweb I  don't see any problems with the datatable. I've just pushed the update to Boostrap 3.4.1, and I've also published a new snapshot on Maven Central. Can you have a look if the glitches you've reported are still there?

mtvweb:
Still there. Just downloaded bootsfaces-1.5.0-20200429.204938-1.jar and modified selectOneMenu.xhtml from BootfacesWeb.xml second b:selectOneMenu to 

`<b:selectOneMenu value="#{selectOneMenuBean.type}" colMd="2" required="true" label="Select your car's type:" labelColMd="2" styleClass="car-class" field-size="sm">`

Attached Screenshots from MacOS Safari, Chrome, Firefox
<img width="463" alt="safari" src="https://user-images.githubusercontent.com/2180107/80649517-80cc5100-8a72-11ea-9b44-b274a513b364.png">
<img width="386" alt="chrome" src="https://user-images.githubusercontent.com/2180107/80649527-845fd800-8a72-11ea-8383-c1b541973d45.png">
<img width="598" alt="firefox" src="https://user-images.githubusercontent.com/2180107/80649535-875ac880-8a72-11ea-8120-6628f56da2f5.png">

mtvweb:
I think Datatables problem is already fixed in master by upgrading datatables.net to a newer version by @geopossachs

stephanrauh:
We've finally managed to solve the showcase bug. Now the datatable looks mostly OK, apart from the first dropdown. Does anyone have an idea why this happens?

mtvweb:
Is the current version only at your showcase?

[...]

PR #1099, opened 2019-10-24
Fix includes for Properties set with a ValueExpression

stjheitc:
if a valueExpression was used for properties that toggles the inclusion of scripts or css if a constant value is passed these inclusion were not added . By adding some additional code to setValueExpression these forces adding the required libraries, even if the final value may be empty or null, which isn't known during set.

Fixes TheCoder4eu#1098
Fixes TheCoder4eu#1052


mtvweb:
Also added new datatable cdn. Fixing the strange filter mentioned in #1093

Fixes #1100

stjheitc:
Fixes also TheCoder4eu/BootsFaces-OSP#1104

mtvweb:
I don‘t think that it will Fix it. as an additional Note. The problem occurs only, if you use only Dynamic Values, i.E. a Page without Layout and a Single commandbutton with an awesome Icon set by el

mtvweb:
Still not working without this pull request. If a page contains only dynamic assignments to awesome icons. It will be not included, because only a static assignment is altering the inclusion of the fontawesome includes.

stephanrauh:
@mtvweb I tried to merge your PR yesterday. However, I'd like to exclude all the JavaScript and CSS files. I don't think they're related to any bug you've fixed. Plus, I've updated to Bootstrap 3.4.1, which probably explains the changes. I'm afraid I'm not a git pro (no pun intended!). Please apologize the inconvenience.

Is it possible to cherry-pick the Java files? Alternatively, can you update the JS and CSS files so I don't see them in the PR?

BTW, is it true you're updating the PR for three or four years now? Keep up the good work - und just ignore my embarrassed face. :)

Best regards,
Stephan

mtvweb:
Just created a new PR, Limited to the relevant files only. So we can remove this PR, without merging it.

stephanrauh:
I'm closing this PR in favor of #1134.

Issue #1100, opened 2019-10-25
Colreorder plugin is missing in DataTables CDN url


mtvweb:
The can url is missing the column reorder plugin.
the cr -part is missing in the cdn DataTables resource url

mtvweb:
Additionally DataTables.min.js is included twice with different cdn values

geopossachs:
hello @mtvweb that's wonderful that you want to support the project.

i would like to review your changes. It is possible to shorten the pull request to the central changes? 
(rebase; chery pick; ...)

mtvweb:
Simply changed the cdn datasource to the current one

`
AddResourcesListener.addDatatablesResourceIfNecessary("https://cdn.datatables.net/v/bs/jszip-2.5.0/dt-1.10.20/af-2.3.4/b-1.6.1/b-colvis-1.6.1/b-html5-1.6.1/b-print-1.6.1/cr-1.5.2/fc-3.3.0/fh-3.1.6/r-2.2.3/rg-1.1.1/rr-1.2.6/sc-2.0.1/sl-1.3.1/datatables.min.css", "css");		AddResourcesListener.addDatatablesResourceIfNecessary("https://cdnjs.cloudflare.com/ajax/libs/pdfmake/0.1.36/pdfmake.min.js", "css");
AddResourcesListener.addDatatablesResourceIfNecessary("https://cdnjs.cloudflare.com/ajax/libs/pdfmake/0.1.36/vfs_fonts.js", "css");		AddResourcesListener.addDatatablesResourceIfNecessary("https://cdn.datatables.net/v/bs/jszip-2.5.0/dt-1.10.20/af-2.3.4/b-1.6.1/b-colvis-1.6.1/b-html5-1.6.1/b-print-1.6.1/cr-1.5.2/fc-3.3.0/fh-3.1.6/r-2.2.3/rg-1.1.1/rr-1.2.6/sc-2.0.1/sl-1.3.1/datatables.min.js", "js");
`

URLs are generated with the download page.

stephanrauh:
@mtvweb @geopossachs I've merged the PR. Would one of you like to add a demo to the BootsFaces showcase? That'd be great!

Best regards,
Stephan

PR #1103, opened 2019-11-10
Issues/1100 missing cdn link

geopossachs:
mtvweb report a mssing plugin inport in #1100


geopossachs:
@stephanrauh do you know why there are on line 102 and 103 an almost same import in the old state? And can i merge this PR?

PR #1106, opened 2019-11-22
Add DataTables "lengthChange" option

sebsoftware:
This adds an attribute for the dataTables "lengthChange" property to the datatable. Offering the possibility to disable "false" the page length menu in the upper right. The default is "true", so it is still compatible with previous releases.


stephanrauh:
Thanks a lot!

PR #1110, opened 2019-12-13
updating tooltip to v3.4.1

neondlh:
trying to fix CVE-2018-14042

### DIFF
--- a/mavenResources/META-INF/resources/bsf/js/tooltip.js
+++ b/mavenResources/META-INF/resources/bsf/js/tooltip.js
@@ -1,34 +1,677 @@
-/*
- ========================================================================
- Bootstrap: tooltip.js v3.3.7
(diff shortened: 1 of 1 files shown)


geopossachs:
i have found the file tooltip.js more the 11 times into the project

neondlh:
I just found just one tooltip.js, can you please point me which other files do I have to update?
Also, I didn't found this js in gradle resources, that was kind of weird.
Thanks!

stephanrauh:
I believe in this particular case, the file is `/build.properties`. As for third party libraries, the root folder is `gradleResources/staticResources`. Riccardo has already updated the ´/build.properties` to version 3.4.1, so I'm going to close the ticket now. 

Even so, thanks a lot!

Best regards,
Stephan

Issue #1097, opened 2019-10-03
selectOneMenu visual problem since 1.4.2


fdzaebel:
Hello together,

since version 1.4.2 of BootsFaces the b:selectOneMenu component is not good readable on Safari browsers. Please see an attached example picture.
<img width="201" alt="Bildschirmfoto 2019-10-03 um 18 08 50" src="https://user-images.githubusercontent.com/46503230/66144294-dfc70600-e608-11e9-93c1-f17bf844cb00.png">

With version 1.4.1 everything was fine.

It would be great if anybody could give me a quick workaround for this.

Thanks a lot!

geopossachs:
Hello, @fdzaebel  thank you for your feedback. 

I believe the bug was coming with the fix for #1090 
You can take a look at the CSS of 'select' and change it. 
If you find a solution we are interested in it too

fdzaebel:
> You can take a look at the CSS of 'select' and change it.
> If you find a solution we are interested in it too

Do you mean this select?

```
select {
        -webkit-appearance: none;

        -moz-appearance: none;

        appearance: none;

        background: url(dropdown.ico) right / 20px no-repeat #fff;

        padding-right: 20px;
}
```

geopossachs:
yes, i mean this section

fdzaebel:
I Am very sorry but I am not a CSS specialist.

But I tried around with BootsFaces "select2=true" option and can say that using this option everything looks ok in IE, Firefox and Safari. 
<img width="201" alt="safari" src="https://user-images.githubusercontent.com/46503230/66149440-6680e080-e613-11e9-90b9-dfd331e29fca.png">
<img width="201" alt="ie" src="https://user-images.githubusercontent.com/46503230/66149441-67197700-e613-11e9-8e69-8772ef7bf176.png">
<img width="201" alt="ff" src="https://user-images.githubusercontent.com/46503230/66149442-67197700-e613-11e9-8dbb-6f3669a75623.png">


Does this help?

sebsoftware:
Upgrading to Bootswatch does not fix this issue.
After setting -webkit-appearance to none the line height is set to 30px by the core.css for a "sm" sized select i.E.. With -webkit-appearance set to the default "menu list-button" it works fine. Another way would be modifying the line-height to individual values in bsf.css.

stephanrauh:
[...]

Issue #1090, opened 2019-08-01
b:selectOneMenu: wrong styling in Firefox


stephanrauh:
This is what it looks like in Chrome:

![image](https://user-images.githubusercontent.com/3045767/62315438-9685f900-b495-11e9-9869-52fcaa3026d0.png)

In Firefox, the dropdown looks a bit odd:

![image](https://user-images.githubusercontent.com/3045767/62315483-ac93b980-b495-11e9-8b45-74ca6741146f.png)

geopossachs:
I could found the same bug in native Bootstrap too

https://getbootstrap.com/docs/3.4/css/#selects

And a solution, if you like 

https://stackoverflow.com/questions/20366166/bootstrap-3-styled-select-dropdown-looks-ugly-in-firefox-on-os-x

https://github.com/twbs/bootstrap/issues/16201#issuecomment-260872271

https://github.com/seansfkelley/synology-download-manager/commit/c97c5b0be3c7bff5a716ee7436cc3918e8ff1852

`
select {
        -webkit-appearance: none;
        -moz-appearance: none;
        appearance: none;
        background: url(dropdown.ico) right / 20px no-repeat #fff;
        padding-right: 20px;
}
`

stephanrauh:
Awesome. I shied away from solving the bug quickly because I believed I had to use the dropdown menu. This solution looks much less disruptive.

stephanrauh:
I've pushed your solution to GitHub.

@geopossachs If you've got some spare time, would you mind to have a quick glance over everything related to `<select>`? Such as the `select2` variant, and `selectMultiMenu`, things like this?

geopossachs:
i maked the bill without the IE 

![grafik](https://user-images.githubusercontent.com/10324450/63384365-f8ec5e00-c39e-11e9-8501-2fa9ff6a6a7d.png)

Is the jQuery plugin bootstrap-select an option?
https://developer.snapappointments.com/bootstrap-select/
https://github.com/snapappointments/bootstrap-select/

geopossachs:
i found in the Stack Overflow thread a tipp to solve the bug - it works :-)

https://developer.mozilla.org/en-US/docs/Web/CSS/::-ms-expand

PR #1119, opened 2020-01-14
Fix CSS for select with other field sizes

sebsoftware:
The select was rendered distorted in Safari, Chrome, Firefox on macos after the fixing #1090. The no longer set -webkit-appearance to menubutton-list is no longer overriding the line-height. This was set wrongly for field sizes other than md. Also long text was overlayed by arrow images.

Fixes TheCoder4eu/BootsFaces-OSP#1097


stephanrauh:
It seems updating to Bootstrap 3.4.1 fixes the problem. Can you please verify if it works on the source I've just pushed to GitHub?

mtvweb:
Still there. Just downloaded bootsfaces-1.5.0-20200429.204938-1.jar and modified selectOneMenu.xhtml from BootfacesWeb.xml second b:selectOneMenu to

<b:selectOneMenu value="#{selectOneMenuBean.type}" colMd="2" required="true" label="Select your car's type:" labelColMd="2" styleClass="car-class" field-size="sm">
chrome
firefox
safari

stephanrauh:
@geopossachs That's precisely what I did. Which browser are you using? Does the error occur without your IE bugfix #1090, too?

This is what it looks like on my Mac:

Brave (based on Chromium):
image

Safari (looks a bit odd on the right-hand side, but doesn't shift the text southward):
image

Firefox:
image

mtvweb:
MacOS 10.14.6 
Safari Version 13.1 (14609.1.20.111.8)
Chrome Version 81.0.4044.129 (Offizieller Build) (64-Bit)
Firefox 74.0 (64Bit)

stephanrauh:
@mtvweb Oops, sorry for confusing you with Georg. Luckily, you've received my answer nonetheless.

I'm confused. We're running almost the same configuration. I'm running a slightly newer OS, but that shouldn't make much of a difference. Probably the bug is on my side, but I can't spot it. Just to be sure: open your developer tools and disable the cache in the network tab (if you haven't done so already).

You can also try the staging area of BootsFaces. It runs the latest changed I've committed. On my machine, it looks like this:

image

mtvweb:
It seems to be that you have modified the b:selectOneMenu with select2 for testing. I've used the second one without select2.

stephanrauh:
That sounds like one of the thing I'm always doing. But in this case, I don't think so. Mind you, after deliberately switching the second select box to "select2" it looks like so:

image

The most prominent difference being the arrow on the right-hand side of the select box.

Confused greetings :)
Stephan

gradleResources/staticResources/css/bsf.css
```diff
@@ -454,6 +454,15 @@ background-image: url("data:image/svg+xml;charset=utf-8,%3Csvg%20version%3D%221.
   appearance: none;
 }
 
+select.input-sm {
+	line-height: 20px;
+}
```
geopossachs, line 459: this is better then line-height: 30px; of core.css 👍

gradleResources/staticResources/css/bsf.css
```diff
@@ -454,6 +454,15 @@ background-image: url("data:image/svg+xml;charset=utf-8,%3Csvg%20version%3D%221.
...
+
+select.input-lg {
+	line-height: 26px;
+}
+
 select.form-control {
 	background-image: url("data:image/svg+xml;charset=utf-8,%3Csvg%20version%3D%221.1%22%20xmlns%3D%22http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%22%20xmlns%3Axlink%3D%22http%3A%2F%2Fwww.w3.org%2F1999%2Fxlink%22%20x%3D%220px%22%20y%3D%220px%22%20fill%3D%22%23555555%22%20%0A%09%20width%3D%2224px%22%20height%3D%2224px%22%20viewBox%3D%22-261%20145.2%2024%2024%22%20style%3D%22enable-background%3Anew%20-261%20145.2%2024%2024%3B%22%20xml%3Aspace%3D%22preserve%22%3E%0A%3Cpath%20d%3D%22M-245.3%2C156.1l-3.6-6.5l-3.7%2C6.5%20M-252.7%2C159l3.7%2C6.5l3.6-6.5%22%2F%3E%0A%3C%2Fsvg%3E");
+  	padding-right: 25px;
```
geopossachs, line 467: 👍

gradleResources/staticResources/css/bsf.css
```diff
@@ -454,6 +454,15 @@ background-image: url("data:image/svg+xml;charset=utf-8,%3Csvg%20version%3D%221.
   appearance: none;
 }
 
+select.input-sm {
+	line-height: 20px;
+}
+
+select.input-lg {
+	line-height: 26px;
+}
```
geopossachs, line 463: this is better then the line-height: 46px; of core.css 👍

Issue #878, opened 2017-10-04
b:datatable sending null for onselect ajax call - Internet Explorer 11


kvj2020:
I have a datatable that sends null to the MyBean.doSelect(MyObj obj) when I use Internet Explorer 11.  It works fine for Chrome and Firefox.  I am using BootsFaces 1.2.0 latest snapshot
```
<b:dataTable id="bpDataTable" value="#{myBean.list}"
	  select="true" ajax="true"  paginated ="false" searching="false"
	  var="myVal" onselect="ajax:myBean.doSelect(myVal)">
```

sometowngeek:
I have the same issue.

Is anyone looking into this?

@stephanrauh I noticed you are active in the issues tab and I thought maybe this is something worth looking into.

sometowngeek:
@stephanrauh requested I open an issue, but since this issue already exists, I figured I should post it as a comment in here instead of opening another issue ticket for the same thing.

This is posted on [StackOverflow - Getting a null value passed to a Java function via BootsFaces datatable](https://stackoverflow.com/q/53233904/3993154).

-----------

I am trying to create a modal to send user-selected requests to an email address; however, I am having trouble getting the user-selected requests. I keep getting the null value passed through `fooBean.setSelected(id)`.

-------

### Versions:
* BootsFaces: 1.3.0
* Java: 1.8.0
* JSF: 2.0
* Browser: Internet Explorer 11.2x

------

### MCVE of thisThing.xhtml:

    <b:dataTable value="#{fooBean.newRequests}"
        var="foo"
        onselect="ajax:fooBean.setSelected(id)"
        ondeselect="ajax:fooBean.setSelected(id)"
        selectedItems="row"
        selection-mode="multiple"
        selectedRows="#{foo.selected}"
        saveState="false">

        <b:dataTableColumn label="Select">
            <b:selectBooleanCheckbox value="#{foo.selected}" />
        </b:dataTableColumn>

        <b:dataTableColumn label="Status" value="#{foo.status}" />

        <b:dataTableColumn label="Request Number"
            value="#{foo.requestNumber}"
            data-type="string" />

        <b:dataTableColumn label="ID" value="#{foo.id}" />

        <b:dataTableColumn value="#{foo.storeName}"
            label="Store Name" />
    </b:dataTable>

-------

### MCVE of fooBean.java:

    @ManagedBean(name="fooBean")
    @ViewScoped
[...]

PR #1121, opened 2020-01-20
#878 fix a diffrent between firefox and internet explorer

geopossachs:


src/main/java/net/bootsfaces/component/dataTable/DataTable.java
```diff
@@ -258,20 +258,23 @@ public void setColumnInfo(List<String> columnInfo) {
...
 		String typeOfSelection = (String) context.getExternalContext().getRequestParameterMap().get("typeOfSelection");
-		context.getELContext().getELResolver().setValue(context.getELContext(), null, "typeOfSelection",
-				typeOfSelection);
+		context.getELContext().getELResolver().setValue(context.getELContext(), null, "typeOfSelection", typeOfSelection);
 		try {
 			int oldIndex = getRowIndex();
+			// after https://datatables.net/reference/event/deselect#Description indexes would be an array
+			// firefox und chrome do not send a array, only the old IE send a array to the server
+			indexes = indexes.replace("[", "").replace("]", "").trim();
 			int index = Integer.valueOf(indexes);
```
kd-apzumi: should not it be an Integer type? it can be null I think if the indexes is not provided

src/main/java/net/bootsfaces/component/dataTable/DataTable.java
```diff
@@ -258,20 +258,23 @@ public void setColumnInfo(List<String> columnInfo) {
...
 			setRowIndex(index);
 			super.queueEvent(event);
 			setRowIndex(oldIndex);
-		} catch (Exception multipleIndexes) {
+		} catch (NumberFormatException multipleIndexes) {
```
kd-apzumi: Previously all type of exceptions were handled, now it is only NumberFormatException, is it enough?

PR #1122, opened 2020-01-22
add a check if a button is used in a grid layout

geopossachs:
i think we need (" btn-block") if the dropButton is used in a grid layout


stephanrauh:
Difficult to say. It's a breaking change. But it's likely that most people expect that behavior.

Is there any antidote if it really breaks the layout of an application?

geopossachs:
my workaround in the current version

  <b:dropButton value="Änderungen speichern"
                                  look=**"success btn-block"**
                                  col-md="2" offset-md="8"
                                  drop="up" >

stephanrauh:
Yeah, sure. Mind you, I've already merged your PR. So I wonder what to say to developers who relied on what we consider a bug. However, I suppose we get away with such a breaking change when publishing a new major version such as 1.5.0. It's not exactly semantic versioning, but it's close. :)

geopossachs:
if the screen is tall, you won't see anything
tall_without_bug

currently, if the screen to small, the button is too large
small_with_button-text_overflow

with the bugfix, if the screen to small, the button is well but it exists a button-text cut
small_with_button-text_cut

Issue #1117, opened 2020-01-13
Adding a Navbar to a stateless page causes a session to be created


exabrial:
Open source work often goes unthanked, so thank you for being great stewards of BootFaces. We love the framework.

I'm not completely convinced this is a bug in BootsFaces, but I figured I'd log a bug anyway. We noticed that adding a Navbar (and likely any bootsfaces components) will cause a session to get created, even on a page with a stateless view. We're running Apache TomEE 7.0.6 which uses OpenWebBeans 1.7.6 and MyFaces 2.2.12. I grabbed a stack trace so you can see how it gets created.

It could be that `CDIManagedBeanHandlerImpl` needs to be a little smarter and not dip down into the session when on a stateless view.

```
SessionContext.<init>() line: 39	
CdiAppContextsService(WebContextsService).initSessionContext(Object) line: 480	
CdiAppContextsService(WebContextsService).startContext(Class<Annotation>, Object) line: 311	
BeginWebBeansListener.sessionCreated(HttpSessionEvent) line: 150	
StandardSession.tellNew() line: 388	
StandardSession.setId(String, boolean) line: 360	
StandardSession.setId(String) line: 341	
StandardManager(ManagerBase).createSession(String) line: 656	
Request.doGetSession(boolean) line: 3087	
Request.getSession(boolean) line: 2476	
RequestFacade.getSession(boolean) line: 896	
CdiAppContextsService(WebContextsService).lazyStartSessionContext(boolean) line: 820	
CdiAppContextsService(WebContextsService).getSessionContext(boolean) line: 737	
CdiAppContextsService(WebContextsService).getCurrentContext(Class<Annotation>) line: 277	
BeanManagerImpl.getContext(Class<Annotation>) line: 291	
SessionScopedBeanInterceptorHandler(NormalScopedBeanInterceptorHandler).getContextualInstance() line: 89	
SessionScopedBeanInterceptorHandler.getContextualInstance() line: 76	
SessionScopedBeanInterceptorHandler(NormalScopedBeanInterceptorHandler).get() line: 71	
ViewScopeBeanHolder$$OwbNormalScopeProxy0.generateUniqueViewScopeId() line: not available [local variables unavailable]	
CDIManagedBeanHandlerImpl.generateViewScopeId(FacesContext) line: 92	
ViewScopeProxyMap.getWrapped() line: 79	
ViewScopeProxyMap.get(Object) line: 119	
[...]

PR #1123, opened 2020-01-31
WIP Issues/1117

geopossachs:
avoid direct access to UIViewRoot.getViewMap() by encapsulation
using stateless memory UIViewRoot.getViewMap() 
use of suitable code suggestions by the netbeans IDE

#1117


src/main/java/net/bootsfaces/listeners/AddResourcesListener.java
```diff
@@ -73,88 +73,138 @@
...
+			return rootView.getAttributes();
+		} else {
+			// if a session exists, we can use the viewMap-store
+			// check if a session exists
+			boolean sessionExists = FacesContext.getCurrentInstance().getExternalContext().getSession(false) != null;
+			if (sessionExists) {
+				LOGGER.log(Level.FINEST, "use viewMap-store");
+				return rootView.getViewMap();
+			} else {
+				// as far as I know, a session is started if there is a UIForm
```
exabrial: I don't think this is true (at least in myfaces 2.3) http://www.icesoft.org/JForum/posts/list/22733.page

PR #1134, opened 2020-04-30
Fix for icons includes assigned by valueExpression (Replacement for 1099)

### LINKED ISSUE #1052: fontawesome only included by static references
If you add components using fontawesome icons, fontawesome will not be included, if the references are not static. This happens if the iconreference is set by an EL expression. Even if you use a composite, awesome isn't recognised to be included by AddResourcesListener.

### LINKED ISSUE #1098: InputText mask library not included with dynamic mask
If the mask is added as a constant it is working, well, but if the mask is taken i.E. from a message than the "js/jquery.inputmask.bundle.min.js" script will not be included.

This works:
`<b:inputText	value="#{schedules.startDate }"
label="${ _msg.startDate }" mask="dd.mm.yyyy"
render-label="true"
span="6" maxlength="20" tabindex="4" required="true" />`

This doesn't work, it it is the only field with a mask within xhtml:

`<b:inputText	value="#{schedules.startDate }"
label="${ _msg.startDate }" mask="${ _common.longDateMask }"
render-label="true"
span="6" maxlength="20" tabindex="4" required="true" />`

mtvweb:
if a valueExpression was used for properties that toggles the inclusion of scripts or css if a constant value is passed these inclusion were not added . By adding some additional code to setValueExpression these forces adding the required libraries, even if the final value may be empty or null, which isn't known during set.

Fixes #1098
Fixes #1052

This is a replacement for PR 1099. Limited to the relevant files only. Hopefully not forgotten anything.

### DIFF
--- a/src/main/java/net/bootsfaces/component/button/Button.java
+++ b/src/main/java/net/bootsfaces/component/button/Button.java
@@ -63,6 +63,11 @@ public class Button extends HtmlOutcomeTargetButton implements IHasTooltip, IRes
 	public void setValueExpression(String name, ValueExpression binding) {
 		name = BsfUtils.snakeCaseToCamelCase(name);
 		super.setValueExpression(name, binding);
+				
+		if ("iconAwesome".equals(name))
+			AddResourcesListener.setNeedsFontsAwesome(this);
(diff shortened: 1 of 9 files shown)


stephanrauh:
Looks fine by me. I'm going to merge the PR now. Thanks at lot for your support and your patience!

Best regards,
Stephan

Issue #1020, opened 2018-10-20
dateTimePicker min/maxDate types are not all supported or wrongly documented


jepsar:
See https://stackoverflow.com/questions/52863344/bootsfaces-typeerror-mindate-could-not-parse-date-parameter

geopossachs:
This will be work:

``
<b:dateTimePicker value="startdate" required="true" minDate="now" colMd="one-fourth" labelColMd="1" label="Start" format="DD-MM-YYYY" />
``

the datepicker comes, i belive, from https://github.com/Eonasdan/bootstrap-datetimepicker/blob/master/src/js/bootstrap-datetimepicker.js and has the fix string _now_ and _moment_ inside or parse the string with moment.js

> When creating a moment from a string, we first check if the string matches known ISO 8601 formats, we then check if the string matches the RFC 2822 Date time format before dropping to the fall back of new Date(string) if a known format is not found.
https://momentjs.com/docs/#/parsing/

This will be not work:
``
<b:dateTimePicker value="startdate" required="true" minDate="#{now}" colMd="one-fourth" labelColMd="1" label="Start" format="DD-MM-YYYY" />
``
-> the Java EL convert {#now} on server side before with a _wrong_ format

geopossachs:
https://stackoverflow.com/questions/9174310/display-current-date-on-jsf-page

>  <p:calendar id="myCalendar" pattern="dd/MM/yyyy" value="#{mybean.myDate}" maxdate="#{now}" />

But we can take a look by the primefaces communtiy :-)

stephanrauh:
I've dropped `#{now}` because it collides with a bean with the same name in OmniFaces (and their `#{now}` is both documented and configurable).

geopossachs:
It works with .toInstant().toString() or with moment.js strings

```
                <b:dateTimePicker value="startdate" required="true" colMd="one-fourth" labelColMd="1" label="Start" format="DD-MM-YYYY" >
                    <!-- <f:attribute name="minDate" value="moment(#{session.lastAccessedTime})"/> -->
                    <!-- <f:attribute name="minDate" value="moment()"/> -->
                    <!-- <f:attribute name="minDate" value="now"/> -->
                    <f:attribute name="minDate" value="#{pageBean.currentDate.toInstant().toString()}" /> 
                </b:dateTimePicker>
```
 i add support for Date()

geopossachs:
[...]

PR #1136, opened 2020-05-01
#1020 add support for Date-Type by min/maxDate in dateTimePicker

geopossachs:


stephanrauh:
Do we have to modify the BootFaces.jsfdsl file, too?

I also wonder if the signature of the setter is OK. What happens when you call `setMaxDate(String)` with a date? Did I miss something?

While you're at it: should we support the new Java8 Date/Time API, too?

geopossachs:
> Do we have to modify the BootFaces.jsfdsl file, too?

It is possible to accept String and Date twice?

> I also wonder if the signature of the setter is OK. What happens when you call `setMaxDate(String)` with a date? Did I miss something?
> 
i have use <f:attribute name="minDate" value="#{register.dateOfBirth}" /> and so i can set a Date() 

> While you're at it: should we support the new Java8 Date/Time API, too?

i would prefer Java8 Date/Time API in the next Version 1.6

stephanrauh:
> It is possible to accept String and Date twice?

That's the domain of languages like TypeScript. :)

I guess it should be possible to define two setters, one for each type. However, there's no easy fix for the getter. Probably it's best to simply use the least common denominator `Object`.

PR #1141, opened 2020-06-02
WIP Issues/995

### LINKED ISSUE #995: input with tags damages the setting process
BootsFaces Version 1.3.0

Assume this simple example:
```xml
<b:form>
  <b:inputText tags="true" value="#{myBean.myValue}" />
  <b:commandButton action="#{myBean.submit}" value="submit" />
</b:form>
```
```java
@ManagedBean
@SessionScoped
public class MyBean {
  private String myValue;
  public String getMyValue(){ return myValue; }
  public String setMyValue(String myValue){ this.myValue = myValue; }
  public String submit() {
    System.out.println(myValue);
  }
}
```
If the `tags`-attribute is set to true, the method `setMyValue` is not used or may be used, but the value does not be submitted. The output is clear for this case.

Removing the `tags`-attribute activates the setter again.

geopossachs:


src/main/java/net/bootsfaces/component/inputText/InputTextRenderer.java
```diff
@@ -260,25 +258,23 @@ public void encodeEnd(FacesContext context, UIComponent component) throws IOExce
...
-		// They initialize tagsinput manually and empty duplicated 'name' attribute
-		// on generated input (which holds the original HTML id).
 		if (inputText.isTags() && (!inputText.isTypeahead())) {
 			String id = fieldId; // input id
 			id = id.replace(":", "\\\\:"); // escape the id for jQuery
 			rw.startElement("script", null);
-			String js = "$('#" + id + "').tagsinput();" + //
-			            "$('#" + id + "').attr('name','');";
+			String js = "$('#" + id + "').tagsinput();";
+			js +=  "$('#" + id + "').siblings('.bootstrap-tagsinput').addClass('form-control');";
```
geopossachs, line 266: with the update to 0.8.0 the bug with the 'name' was fixed, but there is a new bug fix a missing class

src/main/java/net/bootsfaces/component/inputText/InputTextRenderer.java
```diff
@@ -297,27 +294,19 @@ public void encodeEnd(FacesContext context, UIComponent component) throws IOExce
 						"}" + //
 						"});";//
 
-				// The following lines fix issue #1079 on tags with typeahead.
-				// They empty duplicated 'name' attribute on generated input (which holds the original HTML id).
-				String inputId = fieldId; // input id
-				inputId = inputId.replace(":", "\\\\:"); // escape the id for jQuery
-				js += "$('#" + inputId + "').attr('name','');";
+                                js +=  "$('." + escapedClientId + "').siblings('.bootstrap-tagsinput').addClass('form-control');";
```
geopossachs, line 298: with the update to 0.8.0 the bug with the 'name' was fixed, but there is a new bug fix a missing class

gradleResources/staticResources/css/ext/input-tags.css
```diff
@@ -6,7 +6,9 @@
 .bootstrap-tagsinput .twitter-typeahead {
 	width: auto;
 	float: none;
-	position: relative !important;
+	position: absolute !important;
```
geopossachs, line 9: @stephanrauh i have move the css from bootstrap-tagsinput.css to input-tags.css but i'm not sure if we need relative or absolute

PR #1152, opened 2020-08-18
Fix expanded-panels for Accordion inside tab

sebsoftware:
Further development showed up that a panel of an accordion inside a tab gets a  more depth client id. Now it takes the last part of the client id for comparison. Due to the fact that it is required that a panel is the direct child of an accordion it should be the correct id.


stephanrauh:
This makes sense. It seems I can't merge your PR after five years (sorry about that!), so I added a commit.

stephanrauh:
You bug fix is soon going to land with version 2.0.0-SNAPSHOT. See #369 on how to get it.

